PR #9, opened 2021-06-23
Minimum SDK

ndzik:


src/ledger/backend/connection.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+		let testTimeout: NodeJS.Timeout;
+		let cb: Function;
+		const conn = new LedgerAdapter(contract);
+		const depositRegistered = new Promise<boolean>((resolve, reject) => {
+			testTimeout = setTimeout(reject, 10000);
+			cb = () => {
+				resolve(true);
+			};
+			conn.on("Deposited", cb);
```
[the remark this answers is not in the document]
RmbRT: The constructor should then return the `finally(() => { conn.off("deposited", resolveTrueFn) })`, so that the caller does no longer have to care about cleaning up.

src/ledger/backend/connection.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+		assets.addAsset(ETHZERO, amount);
+		const contract = Erdstall__factory.connect(
+			testenv.erdstall,
+			testenv.users[BOB],
+		);
+
+		let testTimeout: NodeJS.Timeout;
+		let cb: Function;
+		const conn = new LedgerAdapter(contract);
+		const depositRegistered = new Promise<boolean>((resolve, reject) => {
```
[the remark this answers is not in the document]
RmbRT: Maybe this is not the best idea for this situation, after all, see below. Keeping it here for you to see as an inspiration for later or something.

src/enclave/provider.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+			case Transfer: {
+				const txc = tx as Transfer;
+				return this.onmessage!(
+					new MessageEvent("erdstall", {
+						data: TypedJSON.stringify(
+							new Result(
+								id,
+								new TxReceipt(
+									tx,
+									new Account(
```
RmbRT, line 77: I think it would be better to create constructors for the message types, such as `NewTxReceiptResult(id, tx, acc) Result` or something. Or, even better, create a constructor that already creates a MessageEvent. Because so many levels of indentation make it hard to understand what is even going on.

src/enclave/connection.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+			// subscribing to exitproofs works.
+			conn.once("exitproof", () => {
+				exitedCalled = true;
+			});
+
+			sendBPtoClient();
+			expect(
+				exitedCalled,
+				"exit callback should not be triggered by normal balanceproofs",
+			).is.false;
```
RmbRT, line 102: wtf. Shouldn't this be a function call?

src/enclave/connection.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+			const exitPromise = conn.exit(pkgtest.NewRandomExitRequest(rng));
+
+			sendBPtoClient(); // balanceProof which should not trigger the exit request.
+			sendEPtoClient();
+
+			const returnedEP = await exitPromise;
+			expect(
+				returnedEP.balance.exit,
+				"the returned exitproof should be a valid exitproof",
+			).to.be.true;
```
RmbRT, line 123: is.true vs to.be.true??? I'm not a big fan of this testing library's API.

RmbRT (changes requested):
`ledger/backend/index.ts` is empty. Delete or set it up properly. (cannot comment in empty files lol)

src/ledger/backend/connection.spec.ts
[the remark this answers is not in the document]
ndzik: I am fond of this idea.

src/enclave/provider.spec.ts
ndzik: No fan of that "arrow-head" code? :laughing: 

I remember being bothered by this too, seems like I forgot to clean this up before handing it in.

src/enclave/connection.spec.ts
ndzik: Not necessarily.

src/enclave/connection.spec.ts
ndzik: I will at least use the same style ;D

RmbRT (approved):
Pretty good

PR #11, opened 2021-07-07
docs: Add README.md

RmbRT:


README.md
```diff
@@ -0,0 +1,134 @@
...
+```ts
+// Initialize these fields with values received from your ethers or web3 setup.
+var ethAccount: {address: SDK.ledger.Address, signer: ethers.Signer};
+// Create client object.
+let Erdstall = SDK.NewClient(
+	ethAccount.address,
+	ethAccount.signer,
+	new URL("https://operator.erdstall.dev"));
+// Set up any event listeners needed by the application.
+Erdstall.once("open", () => { myDApp.run(Erdstall); });
```
matthiasgeihs, line 47: Try to avoid using a string when you can use a constant instead. Otherwise it is easy to introduce bugs when maintaining or changing the code.

You can replace `"open"` by a constant `Erdstall.OpenEvent`.

Even better would be to have a specific handler for each event type: `Erdstall.OnOpen`.

README.md
```diff
@@ -0,0 +1,134 @@
...
+// Establish the client's connection.
+Erdstall.initialize();
+```
+
+**Setting up events**&emsp;<!--
+-->After creation, the Erdstall client is now only partially initialized.
+Now is the time to set up any `SDK.ledger.ErdstallEvent` and `SDK.enclave.EnclaveEvent` handlers using `Erdstall.on()` and `Erdstall.once()`.
+The following `EnclaveEvent` events exist (emitted event types are written in parentheses):
+
+* **`"open"`**:
```
matthiasgeihs, line 57: See comment above. You should at least use constants for the event names. Even better would be to have a separate handler function for each event type. See [jQuery Events](https://api.jquery.com/category/events/), for example.

README.md
```diff
@@ -0,0 +1,134 @@
...
+* `"TokenRegistered"`:
+		a specific token contract has been registered on-chain.
+
+
+**Finishing up**&emsp;<!--
+-->After all events have been set up, call `Erdstall.initialize()` to establish the client's connection to the Erdstall ledger.
+You're now good to go!
+The `open` event will be triggered and your logic will launch.
+
+## Using the Erdstall client
```
matthiasgeihs, line 98: It would be great to add a reference to a minimal working example.

matthiasgeihs (commented):
Very informative, some comments below.

- Try to avoid using strings if you can use constants. This applies to how you define event names. Even better: You can define a separate handler for each event type, see for example [how jQuery does it](https://api.jquery.com/category/events/).
- While this introduction is very helpful already, after reading it, I felt like I want to see some real working code. I think a minimal working examples would be a great addition.

README.md
RmbRT: Typescript, as far as I know, has a compile time check to check the validity of the string.

README.md
ndzik: Typescript has type narrowing, which means strings like that are evaluated at compile time. It is not possible to use this API with non-existent events or similar.

Personally I dislike the usage of these `string`s, but it is common in every other web3-provider/library, so I did not want to break with the flow and keep it as ordinary as possible.

I also plan to use some variation of dependent types such that users of this API, given that they use some form of auto-completion, are given the correct eventhandler signature as soon as an event is specified. But I wanted to do this in the second iteration.

sebastianst: These are issues that concern the sdk in general and not this README PR. I agree that string constants are nicer, and specific methods are best. But let's just get a first version of the README up and move these discussions into separate issues.

README.md
sebastianst: See below comment, I generally agree, bu this is just the README PR.

README.md
sebastianst: created new issue #13

README.md
```diff
@@ -0,0 +1,134 @@
+# Erdstall typescript SDK
```
sebastianst, line 1:
```suggestion
# Erdstall TypeScript SDK
```

README.md
```diff
@@ -0,0 +1,134 @@
...
+
+**What is Erdstall?**&emsp;<!--
+-->[Erdstall](https://erdstall.dev) is a layer-2 platform for Ethereum developed by [PolyCrypt GmbH](https://polycry.pt) based on research by the chair of applied cryptography at the technical university of Darmstadt, Germany.
```
sebastianst, line 4:
```suggestion
-->[Erdstall](https://erdstall.dev) is a layer-2 platform for Ethereum developed by [PolyCrypt](https://polycry.pt) based on research by the chair of applied cryptography at the technical university of Darmstadt, Germany.
```

README.md
```diff
@@ -0,0 +1,134 @@
...
+It uses a [TPM](https://en.wikipedia.org/wiki/Trusted_Platform_Module) — such as Intel SGX or ARM TrustZone — ensuring reliable and trustworthy execution of the platform without the need for public verification of all computations (in comparison to accumulated zero-knowledge proofs, TPMs allow the generation of more compact and less computationally expensive proofs, thereby scaling much better and consuming less power).
+In Erdstall, the TEE operator periodically generates checkpoints, also called balance proofs, that can be used to withdraw funds on-chain.
```
sebastianst, line 6:
```suggestion
In Erdstall, the TEE operator periodically generates checkpoints, also called balance proofs, that can be used to exit the system if the operator were to go offline.
```

README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Getting started
+
+To get your Erdstall app started, there are a few steps you need to take.
+
+**Importing Erdstall**&emsp;<!--
+-->First, import the SDK's main source directory:
+
+```ts
+import * as SDK from "erdstall-ts-sdk/src"
```
sebastianst, line 32:
```suggestion
import * as erdstall from 'erdstall'
```

Please name the package variable `erdstall` and the client instance `client` below. This is more in line with how you'd usually call this. `SDK` is non-descriptive.

sebastianst: I'm about to open a PR that makes this a node package, after which we can use it this way.

README.md
```diff
@@ -0,0 +1,134 @@
...
+It natively integrates with any ethers.js provider, such as Metamask, as its wallet manager.
+The SDK offers full control, but also packages the most important functionality into an easy-to-use API.
+The SDK is designed such that code snippets that use Erdstall can be written generically to be reused within any site using Erdstall.
+That way, common web components, such as payment/donation buttons, can easily be reused on your own homepage.
+
+## Getting started
+
+To get your Erdstall app started, there are a few steps you need to take.
+
+**Importing Erdstall**&emsp;<!--
```
sebastianst, line 28: Add sentence here (will open PR for this soon)
> The Erdstall SDK is not published as a node package yet, so clone it with `git clone https://github.com/perun-network/erdstall-ts-sdk.git erdstall`, then `cd erdstall; yarn link` to locally register the erdstall node module.

README.md
```diff
@@ -0,0 +1,134 @@
...
+-->First, import the SDK's main source directory:
+
+```ts
+import * as SDK from "erdstall-ts-sdk/src"
+```
+
+**Creating the client object**&emsp;<!--
+-->The SDK is straight-forward to use: everything you need is neatly bundled in the `SDK.Erdstall` interface, which you can instantiate via `SDK.NewClient()`, as follows:
```
sebastianst, line 36: rename `SDK` to `erdstall` everywhere

README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```ts
+// Initialize these fields with values received from your ethers or web3 setup.
+var ethAccount: {address: SDK.ledger.Address, signer: ethers.Signer};
+// Create client object.
+let Erdstall = SDK.NewClient(
```
sebastianst, line 42:
```suggestion
let client = SDK.NewClient(
```
or `erdclient`.

sebastianst: In particular, local variables of class instances usually don't start with a capital letter.

README.md
```diff
@@ -0,0 +1,134 @@
...
+```ts
+import * as SDK from "erdstall-ts-sdk/src"
+```
+
+**Creating the client object**&emsp;<!--
+-->The SDK is straight-forward to use: everything you need is neatly bundled in the `SDK.Erdstall` interface, which you can instantiate via `SDK.NewClient()`, as follows:
+
+```ts
+// Initialize these fields with values received from your ethers or web3 setup.
+var ethAccount: {address: SDK.ledger.Address, signer: ethers.Signer};
```
sebastianst, line 40: Do we have a TS type for that?

README.md
```diff
@@ -0,0 +1,134 @@
...
+
+1. The seller creates a trade offer — which is a promise indicating he is willing to trade certain assets for a certain price — and sends it to a market operator.
+2. The market operator broadcasts the trade offer to potential buyers, who can respond with a trade agreement, indicating they are willing to match the offer.
+3. The market operator sends both the trade offer and the trade agreement to Erdstall.
+
+>*This is currently not implemented, we still need to create a more unified API on how to connect the Erdstall client to any Market operator, both for proposing and accepting trades.*
+
+**Off-chain NFT minting**&emsp;<!--
+-->`Erdstall.mint(SDK.ledger.Address, SDK.Uint256) TXReceipt` can be used to create a new token of the specified kind and with the specified ID.
+The NFT can then be traded freely, and is only minted on-chain upon withdrawal.
```
sebastianst, line 134: Please add copyright and license notice to the bottom like in our other repos.

sebastianst (changes requested):

README.md
RmbRT: no. This was just to semantically group those values together while also writing down their types.

RmbRT:
Also updated the sample code to use the new `await client.initialize()` instead of registering an event handler.

sebastianst (approved):

PR #15, opened 2021-07-15
Adapt api to new backend

ndzik:


tsconfig.build.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "include": [
+      "src/**/*",
+      "test/**/*"
```
sebastianst, line 5: I think you can (and should) drop this line. There's already `src/test` and we'll remove the `test` symlink soon.

sebastianst: Same for `tsconfig.test.json`, there you could also drop the `test` include.

And then, you can actually drop the whole `"include"` in `tsconfig.{build,test}.json` because the default `tsconfig.json` already includes only `src/**/*`. You can also remove the empty `"exclude"` from test.json because the main one already has an empty exclude.

src/api/calls/getaccount.ts
```diff
@@ -2,8 +2,8 @@
 "use strict";
 
 import { jsonObject, jsonMember } from "typedjson";
-import { ErdstallObject } from "../object";
-import { Address } from "../../ledger";
+import { ErdstallObject } from "#erdstall/api/object";
+import { Address } from "#erdstall/ledger";
```
sebastianst, line 6: I think this here is a _good_ example for where we want to use package anchor imports: where we have to go into different dirs using `../`s and `../..`s.

src/api/call.ts
```diff
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 "use strict";
 
-import { ErdstallObject } from "./object";
+import { ErdstallObject } from "#erdstall/api/object";
```
sebastianst, line 4: I propose that _within_ a directory, we keep relative imports. This has the advantage that if we were to rename or move the directory they reside in, we don't have to change those imports. This is also more similar to how we can currently do dir-local imports in Go :) In particular, all `export * from './foo'` in `index.ts` files should also be relative.

src/api/object.impl.ts
```diff
@@ -2,10 +2,10 @@
...
 
 import { TypedJSON } from "typedjson";
-import { Transaction } from "./transactions";
-import { ErdstallObject } from "./object";
-import * as calls from "./calls";
-import * as responses from "./responses";
+import { Transaction } from "#erdstall/api/transactions";
+import { ErdstallObject } from "#erdstall/api/object";
+import * as calls from "#erdstall/api/calls";
+import * as responses from "#erdstall/api/responses";
```
sebastianst, line 8: Another example where I'd keep relative imports (gonna stop commenting on these internal imports now :) )

src/erdstall.ts
```diff
@@ -7,16 +7,16 @@
...
+import { TxReceipt } from "#erdstall/api/responses/txreceipt";
+import { BalanceProof } from "#erdstall/api/responses/balanceproof";
+import { Assets } from "#erdstall/ledger";
+import { Address } from "#erdstall/ledger";
+import { Uint256 } from "#erdstall/api/util";
+import { Stages } from "#erdstall/utils";
+import ErdstallEvent from "#erdstall/ledger/event";
+import EnclaveEvent from "#erdstall/enclave/event";
+import Client from "#erdstall/client";
+import { Enclave, EnclaveWSProvider } from "#erdstall/enclave";
```
sebastianst: IMO these are _good_ because we import from independent packages 👍🏻

src/ledger/backend/contracts_deposit.ts
```diff
@@ -3,16 +3,16 @@
...
-import { Address } from "../";
-import { Asset } from "../assets";
-import { Amount } from "../assets";
-import { Tokens } from "../assets";
-import { DepositCalls } from "./tokenmanager";
+import { ETHHolder__factory } from "#erdstall/ledger/backend/contracts/factories/ETHHolder__factory";
+import { IERC20__factory } from "#erdstall/ledger/backend/contracts/factories/IERC20__factory";
+import { IERC721__factory } from "#erdstall/ledger/backend/contracts/factories/IERC721__factory";
+import { ERC20Holder__factory } from "#erdstall/ledger/backend/contracts/factories/ERC20Holder__factory";
+import { ERC721Holder__factory } from "#erdstall/ledger/backend/contracts/factories/ERC721Holder__factory";
```
sebastianst, line 10: this is a special case... if the sub-dir structure `./contracts/factories/*` is tightly coupled to this directory `backend`, I'd also keep local imports, because then the dir-tree `backend/**` is then one coherent module. If something is not coupled, I'd use the `#erdstall` anchor.

src/test/address.ts
```diff
@@ -2,7 +2,7 @@
 "use strict";
 
 import { Address } from "#erdstall/ledger";
-import PRNG from "./random";
+import PRNG from "#erdstall/test/random";
```
sebastianst, line 5: if you want, we could also create a second anchor `#test` -> `src/test`, so a shortcut for `#erdstall/test`. But up to you.

sebastianst (changes requested):

sebastianst:
I'd also rename the commit ` src: Use ambient paths everywhere ` to something like `Use package anchor imports` because they're _not_ ambient imports but internal package imports using a `#erdstall` anchor.

src/api/call.ts
ndzik: Agree.

sebastianst (approved):

PR #17, opened 2021-07-16
Add Method Address.isZero(): boolean

sebastianst:
- Improve .gitignore, add yarn.lock
- Configure exports of subdirs (node and ts)
- node: Update yarn.lock
- ledger: Add Address.isZero()

### DIFF
--- a/src/ledger/address.ts
+++ b/src/ledger/address.ts
@@ -31,8 +31,13 @@ export class Address implements ABIValue {
 	asABI(): any {
 		return this.toString();
 	}
+
 	ABIType(): string {
 		return "address";
+	}
+
+	isZero(): boolean {
+		return this.value.find((v) => v > 0) === undefined;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/ledger/address.ts
```diff
@@ -31,9 +31,14 @@ export class Address implements ABIValue {
 	asABI(): any {
 		return this.toString();
 	}
+
 	ABIType(): string {
 		return "address";
 	}
+
+  isZero(): boolean {
+    return this.value.find(v => v>0) == undefined;
```
ndzik, line 40: Although it works here, it is generally advised to always use the strict `===` comparison operator to avoid unexpected outcomes because of implicit typeconversions.

On another note, are those spaces or tabs used? We have a `.prettier` and `.eslintrc` in the repository which can be used with an autoformatter to avoid this (:

ndzik (changes requested):

src/ledger/address.ts
sebastianst: sorry, using prettier now.

ndzik (approved):
Glorious. :raised_hands:

PR #25, opened 2021-07-23
Add equals methods to Address and BigInteger

sebastianst:
- ledger: Use .every in Address.isZero() implementation
- ledger: Add Address.equals() method
- api/util: Add BigInteger.equals() method

### DIFF
--- a/src/ledger/address.ts
+++ b/src/ledger/address.ts
@@ -39,6 +39,11 @@ export class Address implements ABIValue {
 	isZero(): boolean {
 		return this.value.every((x) => x === 0);
 	}
+
+	equals(other: Address): boolean {
+		return (this.value.length === other.value.length)
+			&& this.value.every((x, i) => x === other.value[i]);
+	}
 }
 
 CustomJSON(Address);
--- a/src/ledger/address.ts
+++ b/src/ledger/address.ts
@@ -37,7 +37,7 @@ export class Address implements ABIValue {
 	}
 
 	isZero(): boolean {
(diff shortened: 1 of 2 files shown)


src/ledger/address.ts
```diff
@@ -39,6 +39,11 @@ export class Address implements ABIValue {
 	isZero(): boolean {
 		return this.value.every((x) => x === 0);
 	}
+
+	equals(other: Address): boolean {
+		return (this.value.length === other.value.length)
```
ndzik, line 44: `( ... )` are not needed here.

Up to you if you leave them or not, I just wanted to comment something here because I like the use of `.every(...)` :D

ndzik (commented):
I wonder whether we should add an `.equals(...)` Method to the other base types of our SDK?

Like for `{Asset|Assets}` to keep an intuition for users of our SDK.

Then `Asset`, being an abstract class, would simply implement it using the existing `cmp()` methods.

```typescript
class Asset {
// ....
    equals(v: Asset): boolean {
        return this.cmp(v) === "eq";
    }
// ....
}
```

Same for `Assets`.

We can also just create an extra issue for that.

src/ledger/address.ts
sebastianst: I prefer to leave as is because it separates more clearly from the next boolean expression.

sebastianst:
If we need the equals methods for `Asset` types, we can add it. But for now, I'd just put it in a separate issue.

ndzik (approved):
Magnum opus :muscle:

PR #28, opened 2021-07-26
Change Asset.fromJSON impl. to use dep. inj.

sebastianst:
- ledger/assets: Change Asset.fromJSON impl. to use dep. inj.
- Remove re-exports from top-level index
- Consistently import Assets from ledger/assets package


sebastianst:
Tests run on 2. commit, but not on 3rd. Investigating...

ndzik (approved):
What a natural :muscle:

PR #34, opened 2021-07-28
Integration test + Trades

RmbRT:
To run the test:
environment variable `ERDSTALL_EXE` must point to `cmd/walkthrough`'s main executable in branch `walkthrough-operator`.
Then, just run `make integration`.


src/client.ts
```diff
@@ -140,7 +140,7 @@ export default class Client implements Erdstall {
 
 		const exittx = new ExitRequest(this.address, await this.nextNonce());
 		await exittx.sign(this.erdstallConn.erdstall(), this.signer);
-		return this.enclaveConn.exit(exittx);
+		return await this.enclaveConn.exit(exittx);
```
ndzik, line 143: Any special reason for `await`ing every Promise?

Less performant in production vs just returning, but makes debugging a little bit easier AFAIK. Just curious.

src/enclave/connection.ts
```diff
@@ -91,10 +92,14 @@ export class Enclave implements EnclaveConnection {
...
 
+	public async trade(tx: Trade): Promise<TxReceipt> {
+		return this.sendCall(tx) as Promise<TxReceipt>;
+	}
+
 	public async exit(exitRequest: ExitRequest): Promise<BalanceProof> {
 		const p = new Promise<BalanceProof>((resolve, reject) => {
 			this.once("exitproof", resolve);
-			this.sendCall(exitRequest).catch(reject);
+			this.sendCall(exitRequest).catch(err => reject(new Error(err)));
```
ndzik, line 102: `err` is not guaranteed to be a  `string`. Shouldn't this die when an object is fed to the `Error` constructor, since it is defined as:

```typescript
interface ErrorConstructor {
    new(message?: string): Error;
    (message?: string): Error;
    readonly prototype: Error;
}
```

I think a `instanceof` or similar is needed here. I do not know how this behaves on user-side.

ndzik: Same for every other occurrence of `new Error(...)`.

src/ledger/backend/connection.ts
```diff
@@ -86,6 +86,10 @@ export class LedgerAdapter implements LedgerConnection {
 		assets: Assets,
 	): Promise<Stages<Promise<ethers.ContractTransaction>>> {
 		const calls: DepositCalls = [];
+
+		if(!assets.values.size)
+			throw new Error("attempting to depost nothing");
```
ndzik, line 91: `"attempting to deposIt nothing"`

ndzik (changes requested):

src/client.ts
RmbRT: Just wanted to make sure that `await client.X()` only really returns when `X` finishes, and not before that (ignoring the special case of on-chain calls).

src/api/sig.spec.ts
```diff
@@ -20,4 +20,13 @@ describe("Signatures", () => {
 
 		expect(tx.verify(contract)).to.be.true;
 	});
+
+	it("works for mints", async () => {
+		const rng = test.NewPrng();
```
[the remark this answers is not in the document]
sebastianst: `rng` should be defined in the top level `describe` 👍🏻

src/enclave/connection.ts
sebastianst: so maybe keep as it was before? why the change @RmbRT ?

src/client.ts
```diff
@@ -105,7 +105,7 @@ export default class Client implements Erdstall {
 			assets,
 		);
 		await tx.sign(this.erdstallConn.erdstall(), this.signer);
-		return this.enclaveConn.transfer(tx);
+		return await this.enclaveConn.transfer(tx);
```
sebastianst, line 108: Why all those changes to awaits? If a function returns a `Promise` by its signature, better to return a Promise instead of resolving locally. If it's not necessary, please revert all added `awaits`.

src/enclave/connection.ts
RmbRT: during testing, I had errors, where promises were rejected with non-`Error` objects, and typescript requested me kindly to please only throw `Error`s.

src/client.ts
RmbRT: I had some bugs when transacting or something that were the result of a missing await in an async function. So for good measure, I put awaits everywhere. In my opinion, an async function should always be called with `await`, unless it is intended to be executed in the background.

sebastianst (changes requested):

src/enclave/connection.ts
sebastianst: ok then do an `instanceof Error` check like `return e instanceof Error ? e : new Error(e)` and hope that it's a string ^^

e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Assets } from "#erdstall/ledger/assets";
+import * as assets from "#erdstall/ledger/assets";
+import * as sdk from "#erdstall";
+import { Erdstall } from "#erdstall";
+
+import { ethers } from "ethers";
+import * as fs from "fs";
+import { exec } from "child_process";
+
+(global as any).WebSocket = require("ws");
```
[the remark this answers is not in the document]
RmbRT: Not that I know of. This was the recommended go-to solution, any way.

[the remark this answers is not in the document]
ndzik: No, we were crying.
Imported a bunch of libs to no avail. This was the only way, even with `jsdom-global` as an extension to mocha, there is no default `Websocket` available.

Although one solution would be to always depend on a websocket library in all our code. Then this would be superfluous, but then again... Why should we be forced to do this?

e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+		const user = ethers.Wallet.fromMnemonic(mnemonic, derivationPath);
+		const userAddr = Address.fromString(user.address);
+		return sdk.NewClient(userAddr, user.connect(provider), opAddr);
+	}
+
+	let alice: Erdstall | null = null;
+	let bob: Erdstall | null = null;
+	let charlie: Erdstall | null = null;
+	let dagobert: Erdstall | null = null;
+	let charlieNft: bigint | null = null;
```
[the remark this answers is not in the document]
RmbRT: The problem is that in `let x?: T` the `?` is invalid, and in `let x: T?`, the `?` seems to have a completely different meaning. And I can't use `describe()`-wide direct accessors without again using Maybe-values, which for some reason don't really work for local variables. But I think the array-version should still be helpful.

e2e/sdk_remote_integration.spec.ts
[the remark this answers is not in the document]
sebastianst: OK I see, reality strikes again. Then it's fine 🤡

e2e/sdk_remote_integration.spec.ts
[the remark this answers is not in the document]
sebastianst: what about just `let x: T;`?

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	function makeClient(index: number): Erdstall {
+		const provider = new ethers.providers.JsonRpcProvider(`http://localhost:${nodePort}`);
+		const derivationPath = `m/44'/60'/0'/0/${index+2}`;
+		const user = ethers.Wallet.fromMnemonic(mnemonic, derivationPath);
+		const userAddr = Address.fromString(user.address);
+		return sdk.NewClient(userAddr, user.connect(provider), opAddr);
+	}
+
+	let clients: Erdstall[] = [];
```
sebastianst, line 50: can this even be a `const`?

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+			ResponseDuration: 3,
+			PowDepth: 0,
+			RPCPort: opPort,
+			RPCHost: "0.0.0.0",
+			NodeReqTimeout: 10,
+			WaitMinedTimeout: 60
+		}), accept));
+
+		exec(`$ERDSTALL_EXE -config operator.cfg -epochs 10 -ledger-port ${nodePort} &`);
+		return new Promise(accept => setTimeout(accept, 10000));
```
sebastianst, line 39: So this just returns a promise that resolves after 10 sec?

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+			EpochDuration: 2,
+			ResponseDuration: 3,
+			PowDepth: 0,
+			RPCPort: opPort,
+			RPCHost: "0.0.0.0",
+			NodeReqTimeout: 10,
+			WaitMinedTimeout: 60
+		}), accept));
+
+		exec(`$ERDSTALL_EXE -config operator.cfg -epochs 10 -ledger-port ${nodePort} &`);
```
sebastianst, line 38: Shouldn't you also cleanup the operator in an `after` statement? How do you make sure that this operator is stopped after the test ends or fails?

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+		nft = 89430923n;
+		return clients[2].mint(clients[2].address, nft);
+	});
+
+	it("lets users trade", async () => {
+		// Charlie trades his NFT for Dagobert's ETH.
+		const offer = new Assets();
+		offer.addAsset(clients[2].address, new assets.Tokens([nft]));
+		const expect = new Assets();
+		expect.addAsset(assets.ETHZERO, new assets.Amount(1000n));
```
sebastianst, line 115: The `Assets` constructor now accepts a variadic list of `{address, Asset}`s that are added to it.

src/api/transactions/trade.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Signer, utils } from "ethers";
+
+@jsonObject
+export class TradeFees {
+	constructor(market: Address, fee: assets.Assets) {
+		this.market = market;
+		this.fee = fee;
+	}
+	@jsonMember(Address) market: Address;
+	@jsonMember(()=>assets.Assets) fee: assets.Assets;
```
sebastianst, line 19: members are usually specified before the ctor

src/api/transactions/trade.ts
```diff
@@ -0,0 +1,104 @@
...
+		owner: Address,
+		offer: assets.Assets,
+		request: assets.Assets)
+	{
+		this.owner = owner;
+		this.offer = offer;
+		this.request = request;
+		this.expiry = new BigInteger(2n**64n - 1n); // For now, never expire.
+		this.fees = undefined; // For now, ignore fees.
+		this.sig = undefined;
```
sebastianst, line 47: I think you can just remove those two lines and these fields will be `undefined`.

src/api/transactions/trade.ts
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	public txType() {
+		return Trade;
+	}
+	protected txTypeName(): string {
+		return tradeTypeName;
+	}
+	protected encodeABI(e: ABIEncoder, contract: Address): string {
+		e.encodeTagged(contract, this.offer.sig!, ["bytes", utils.hexlify(this.offer.asABITagged(contract).bytes)]);
```
sebastianst, line 99: if we assert here that field `sig` is always defined, shouldn't we check for that in the ctor?

src/api/transactions/transaction.spec.ts
```diff
@@ -24,6 +29,27 @@ describe("Wiremessages", () => {
 	it("de-/encodes mints", () => {
 		genericJSONTest(testMint, Call);
 	});
+	it("de-/encodes trades", () => {
+		genericJSONTest(testTrade, Call);
+	});
+	it("trade ABI test (fees)", () => {
+		const contract = Address.fromString("0x6e2FD8A6FB6caf3612967AadC749763e9657e9D1");
+		const tradeCall = TypedJSON.parse(testTrade, Call)!;
+		let tradeTx = tradeCall.data as Transaction as Trade;
```
sebastianst, line 38: const

src/api/transactions/transaction.spec.ts
```diff
@@ -24,6 +29,27 @@ describe("Wiremessages", () => {
...
+		expect(tradeTx.offer.fees).to.exist;
+		let offer = tradeTx.offer;
```
sebastianst, line 40: const

src/api/transactions/transaction.spec.ts
```diff
@@ -24,6 +29,27 @@ describe("Wiremessages", () => {
...
...86e6c6da8d56cf1812d048e5a39f3db714c2a195b042d000000000000000000000000000000000000000000000000000000000000022000000000000000000000000000000000000000000000000000000000000000800000000000000000000000006e2fd8a6fb6caf3612967aadc749763e9657e9d100000000000000000000000049ed509ee584a17987156e276f180b234e79292c00000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000114572647374616c6c5472616465466565730000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000035b7ef7803eb1a954517ee58aa9ea5964afd44ce000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000203aeb4f544769e0a8dd5ab3d2a8569336cadeb54db0af60cf79c5c13f8b1c1b86000000000000000000000000ccf0fbe58471cb5e6e4bf25fddb8f8709b24f1fc00000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000020118d5931da69e0069fcb94ef71e248bb22379080bd8983dda622c7c9a59416df");
```
sebastianst, line 43: Where did you derive the golden encodings from? Please add a comment to document how they were derived in the first place.

src/api/transactions/transaction.spec.ts
```diff
@@ -24,6 +29,27 @@ describe("Wiremessages", () => {
...
+		expect(tradeTx.offer.fees).to.not.exist;
```
sebastianst, line 49: Being more specific here with `to.be.undefined` is [recommended](https://www.chaijs.com/api/bdd/#method_exist) and more precise.

sebastianst (changes requested):

src/e2e/sdk_remote_integration.spec.ts
RmbRT: yes, I need to wait for the blockchain etc. to start properly and open its ports.

src/e2e/sdk_remote_integration.spec.ts
RmbRT: The operator terminates after 10 epochs.

RmbRT: And I have no idea how to terminate the process manually.

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+describe("Erdstall-TS-SDK", () => {
+	const opPort = 1433;
+	const opAddr = new URL(`ws://127.0.0.1:${opPort}/ws`);
+	const nodePort = 1362;
+	const mnemonic = "pistol kiwi shrug future ozone ostrich match remove crucial oblige cream critic";
+	let erdstallProcessTerminate: Promise<void>;
+
+	before(async () => {
+		await new Promise(accept => fs.writeFile("operator.cfg", JSON.stringify({
```
sebastianst, line 26: `fs.writeFileSync`

src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+			erdstallProcessTerminate = new Promise<void>((accept, reject) => {
+				process.on("close", accept);
+				process.on("error", reject);
+			});
+			const timeout = setTimeout(accept, 10000)
+			process.on("error", (e: Error) => {
+				clearTimeout(timeout);
+				reject(e);
+			});
+		});
```
sebastianst, line 55: I hope you know what you're doing here 😂

sebastianst (changes requested):

src/e2e/sdk_remote_integration.spec.ts
RmbRT: The setup function's promise waits for 10s or until the operator terminates in failure (or fails to execute). Additionally, another promise is created that waits for termination of the process, without time limit. The process has to be started inside the promise constructor to make the `reject` function visible to the handler I pass to the `exec` call (`on("error")` only seems to trigger if the process was also started correctly, so it cannot catch process creation failures).

sebastianst (approved):

PR #36, opened 2021-07-29
test: Add MockWatcher

sebastianst:
Only supports TxReceipt events for now.


ndzik (approved):
:fire: **HOT** :fire:

PR #37, opened 2021-07-29
Split Client/Session, Reader/Writer

RmbRT:
Current Changes:
- Session can be created from client, currently reuses the same
  websocket connection, so it would be shared among instances. Sessions
  should re-expose everything from Clients (including unbound
  subscriptions). Sessions also share their event handlers with clients.
- Diversified subscription interfaces, now there are Subscriber
  (subscribeFor(Address), subscribeAll()), and BoundSubscriber
  (subscribe()).

TODO:
- Split ledger connection into LedgerReader/LedgerWriter.
- Figure out whether to share websocket connections or not.
- Figure out: should sessions have completely separate event tracking?
- Figure out: should a single client be used to host multiple
  *independent* sessions, or at most one?

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,39 +1,35 @@
 // SPDX-License-Identifier: Apache-2.0
 "use strict";
 
-import { Signer } from "ethers";
-import { ethers } from "ethers";
 
-import { Erdstall } from "erdstall";
+import { ErdstallClient } from "erdstall";
 import { Erdstall__factory } from "#erdstall/ledger/backend/contracts";
-import { TxReceipt } from "#erdstall/api/responses";
-import { BalanceProof } from "#erdstall/api/responses";
 import { ClientConfig } from "#erdstall/api/responses";
-import { Transfer, Mint, ExitRequest, TradeOffer, Trade } from "#erdstall/api/transactions";
-import { EnclaveConnection, EnclaveEvent } from "#erdstall/enclave";
-import { LedgerConnection, LedgerAdapter } from "#erdstall/ledger";
-import { ErdstallEvent, isLedgerEvent } from "#erdstall/ledger";
-import { Assets } from "#erdstall/ledger/assets";
-import { Address } from "#erdstall/ledger";
-import { Uint256 } from "#erdstall/api/util";
-import { EventCache, OneShotEventCache, Stages } from "#erdstall/utils";
+import { Enclave, EnclaveReader, EnclaveEvent } from "#erdstall/enclave";
+import {
+	LedgerWriteConn,
+	LedgerReader,
+	LedgerWriter,
+	Address,
+	ErdstallEvent,
+	isLedgerEvent
+} from "#erdstall/ledger";
+import { EventCache, OneShotEventCache } from "#erdstall/utils";
+import { ethers, Signer } from "ethers";
 
-export const ErrUnitialisedClient = new Error("client unitialised");
-
-export default class Client implements Erdstall {
-	readonly address: Address;
-	private nonce: bigint;
-	private enclaveConn: EnclaveConnection;
-	private erdstallConn?: LedgerConnection;
+export class Client implements ErdstallClient {
+	protected enclaveConn: EnclaveReader;
+	protected provider: ethers.providers.Provider | Signer;
+	protected erdstallConn?: LedgerReader | LedgerWriter;
 	private erdstallEventHandlerCache: EventCache<ErdstallEvent>;
 	private erdstallOneShotHandlerCache: OneShotEventCache<ErdstallEvent>;
-	private signer: Signer;
 
-	constructor(address: Address, signer: Signer, encConn: EnclaveConnection) {
-		this.address = address;
-		this.signer = signer;
-		this.nonce = 1n;
-		this.enclaveConn = encConn;
+	constructor(provider: ethers.providers.Provider | Signer, encConn: EnclaveReader | URL) {
+		this.provider = provider;
+		if(encConn! instanceof URL)
+			this.enclaveConn = Enclave.dial(encConn! as URL);
+		else
+			this.enclaveConn = encConn! as EnclaveReader;
 		this.erdstallEventHandlerCache = new EventCache<ErdstallEvent>();
 		this.erdstallOneShotHandlerCache = new OneShotEventCache<ErdstallEvent>();
 	}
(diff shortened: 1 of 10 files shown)


src/session.ts
```diff
@@ -0,0 +1,160 @@
...
+import { LedgerConnection, LedgerAdapter } from "#erdstall/ledger";
+import { ErdstallEvent } from "#erdstall/ledger";
+import { Assets } from "#erdstall/ledger/assets";
+import { Address } from "#erdstall/ledger";
+import { Uint256 } from "#erdstall/api/util";
+import { Stages } from "#erdstall/utils";
+
+export const ErrUnitialisedClient = new Error("client unitialised");
+
+export default class Session implements ErdstallSession {
```
sebastianst, line 23: I think you can just `extend ErdstallClient`, then you don't need to duplicate and forward most of the calls. Also makes sense from a modelling point of view.

src/erdstall.ts
```diff
@@ -62,33 +65,62 @@ export interface Leaver extends Exiter, Withdrawer {
 	leave(): Promise<Stages<Promise<ethers.ContractTransaction>>>;
 }
 
-export interface Subscriber {
+export interface BoundSubscriber {
 	subscribe(): Promise<void>;
+}
```
sebastianst, line 70: I'd set this to
```suggestion
export interface BoundSubscriber {
	subscribeSelf(): Promise<void>;
}
```
and let the `Subscriber` interface be `subscribe(who?: Address): Promise<void>`. Makes things simpler.

src/erdstall.ts
```diff
@@ -62,33 +65,62 @@ export interface Leaver extends Exiter, Withdrawer {
...
 		Withdrawer,
 		Exiter,
-		Subscriber,
+		BoundSubscriber,
 		Leaver {
 	readonly address: Address;
 	initialize(): Promise<void>;
 }
 
 export function NewClient(
```
sebastianst, line 103: Make alternative constructors static functions on the class they create. And it's convention to start all functions and methods with a small letter.

src/session.ts
```diff
@@ -0,0 +1,160 @@
...
+	readonly address: Address;
+	private client: ErdstallClient;
```
sebastianst, line 25:
```suggestion
	private readonly client: ErdstallClient;
```

src/session.ts
```diff
@@ -0,0 +1,160 @@
...
+	private nonce: bigint;
+	private enclaveConn: EnclaveWriter;
```
sebastianst, line 27:
```suggestion
	private readonly enclaveConn: EnclaveWriter;
```

src/session.ts
```diff
@@ -0,0 +1,160 @@
...
+	private erdstallConn?: LedgerConnection;
+	private signer: Signer;
```
sebastianst, line 29:
```suggestion
	private readonly signer: Signer;
```

src/session.ts
```diff
@@ -0,0 +1,160 @@
...
+import { BalanceProof } from "#erdstall/api/responses";
+import { ClientConfig } from "#erdstall/api/responses";
+import { Transfer, Mint, ExitRequest, TradeOffer, Trade } from "#erdstall/api/transactions";
+import { EnclaveWriter, EnclaveEvent } from "#erdstall/enclave";
+import { LedgerConnection, LedgerAdapter } from "#erdstall/ledger";
+import { ErdstallEvent } from "#erdstall/ledger";
+import { Assets } from "#erdstall/ledger/assets";
+import { Address } from "#erdstall/ledger";
+import { Uint256 } from "#erdstall/api/util";
+import { Stages } from "#erdstall/utils";
```
sebastianst, line 19: group imports from same module

sebastianst (changes requested):

src/session.ts
sebastianst, line 23: Could we let it `extend Client`? Then we avoid redirection of calls and, I think, it makes sense from a modelling point of view that our default implementation of the `Session` uses the concrete default impl of `Client`. Cannot think of a use case right now where we would need an abstract `Client` and use it in this `Session`.

src/session.ts
```diff
@@ -0,0 +1,152 @@
...
+	readonly address: Address;
+	private nonce: bigint;
+	private readonly enclaveWriter: EnclaveWriter;
+	private readonly signer: Signer;
+
+
+	static dial(
+		address: Address,
+		signer: Signer,
+		operatorAddress: URL
```
sebastianst, line 28: please rename to
```suggestion
		operatorUrl: URL
```
it's confusing to use `address` because we usually use it for blockchain addresses.

src/client.ts
```diff
@@ -1,41 +1,51 @@
...
-		this.erdstallEventHandlerCache = new EventCache<ErdstallEvent>();
-		this.erdstallOneShotHandlerCache = new OneShotEventCache<ErdstallEvent>();
+	// Takes (Client) or (Provider | Signer, EnclaveReader | EnclaveWriter | URL).
+	constructor(clientOrLedgerProvider: Client | ethers.providers.Provider | Signer, encConn?: EnclaveReader | EnclaveWriter | URL) {
+		if(clientOrLedgerProvider instanceof Client) {
+			this.provider = clientOrLedgerProvider.provider;
+			this.enclaveConn = clientOrLedgerProvider.enclaveConn;
+			this.erdstallEventHandlerCache = clientOrLedgerProvider.erdstallEventHandlerCache;
+			this.erdstallOneShotHandlerCache = clientOrLedgerProvider.erdstallOneShotHandlerCache;
+		} else {
```
sebastianst, line 40: Wouldn't it be possible to remove the constructor taking a `Client`? And the `Session` constructor should then receive `(address: Address, singer: Signer, encConn: EnclaveWriter | URL)`. I think you'd also have better type safety this way and less type checks.

IMO you could also remove the `| URL` option from the constructors and then pass the `URL` to the `.dial` statics (could create one for Client).

src/session.ts
```diff
@@ -0,0 +1,152 @@
...
+	): Session {
+		return new Session(address, new Client(signer, operatorAddress));
+	}
+
+	// Consumes `client`. `signer` is only needed if the client has no signer.
+	// `client` must have an EnclaveWriter.
+	constructor(address: Address, client: Client, signer?: Signer) {
```
sebastianst, line 35: If we change the constructor here, as I suggested above, to `(address: Address, signer: Signer, encConn: EnclaveWriter)`, we can just call `super(signer, encConn);` and have much less logic, no?

src/session.ts
```diff
@@ -0,0 +1,152 @@
...
+	private readonly enclaveWriter: EnclaveWriter;
+	private readonly signer: Signer;
+
+
+	static dial(
+		address: Address,
+		signer: Signer,
+		operatorAddress: URL
+	): Session {
+		return new Session(address, new Client(signer, operatorAddress));
```
sebastianst, line 30: This would become sth like
```suggestion
  const encConn = new Enclave(new EnclaveWSProvider(operatorUrl));
		return new Session(address, signer, encConn);
```

src/session.ts
```diff
@@ -0,0 +1,152 @@
...
+
+		if(signer)
+			this.signer = signer;
+		else if(this.provider instanceof Signer)
+			this.signer = this.provider as Signer;
+		else
+			throw new Error("client provider cannot be promoted to Signer, no Signer provided");
+		this.provider = this.signer;
+
+		this.enclaveWriter.once("error", () => { this.updateNonce(); });
```
sebastianst, line 53: what is this?

src/session.ts
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	private async nextNonce(): Promise<bigint> {
+		if (this.nonce === 1n) {
```
sebastianst, line 57: why don't we start at 0?

src/erdstall.ts
```diff
@@ -62,33 +63,34 @@ export interface Leaver extends Exiter, Withdrawer {
...
+
+export interface Onboarder {
 	onboard(): Promise<void>;
 }
 
-export interface Erdstall
+export interface ErdstallClient
 	extends Watcher,
+		Subscriber {
+	initialize(): Promise<void>;
```
sebastianst, line 81: Please add comment when `initialize` should be called. Is it before or after `subscribe`?

src/client.ts
sebastianst: And I think you can also omit `| EnclaveWriter` because it extends a reader. So it should be able to change this ctor to `(provider: ethers.providers.Provider | Signer, encConn: EnclaveReader)`.

src/erdstall.ts
```diff
@@ -62,33 +63,34 @@ export interface Leaver extends Exiter, Withdrawer {
 	leave(): Promise<Stages<Promise<ethers.ContractTransaction>>>;
 }
 
+export interface BoundSubscriber {
```
sebastianst, line 66: Please rename to `SelfSubscriber`

sebastianst (changes requested):

src/session.ts
RmbRT: Great catch. I changed it to start as 0, and also fixed a potential race in the updateNonce function when sending multiple transactions concurrently.

src/session.ts
RmbRT: Clarified with a comment.

src/session.ts
RmbRT: The whole constructor is no longer necessary.

sebastianst (approved):

PR #41, opened 2021-08-02
MockWatcher.trade and MockClient

sebastianst:
- utils/arrays: Add equalArray function to new package utils/arrays
- ledger: Use equalArray in Address.equals
- test: Allow optional TradeOffer in newRandomTrade
- test: Add MockWatcher.trade to inject Trade transactions
- !fixup
- test: Add MockClient (extends MockWatcher)


src/utils/arrays/index.ts
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+
+export type TypedArray = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array;
+
+export function equalArray(x: TypedArray, y: TypedArray): boolean {
+	return (x.length == y.length)
```
ndzik, line 6:
```suggestion
	return (x.length === y.length)
```

src/utils/arrays/index.ts
```diff
@@ -0,0 +1,8 @@
...
+		&& (x.every((a, i) => y[i] == a));
```
ndzik, line 7: Considering the function signature this is correct, because `x` and `y` can be of different `TypedArray` types, which would require a comparison function with type coercion.

If this was not intentional I would change the implementation to

```typescript
export functino equalArray<T extends TypedArray>(x: T, y: T): boolean { ... }
```

And 
```suggestion
		&& (x.every((a, i) => y[i] === a));
```

I did not test it, but I would not be surprised if comparing certain edge cases of `Int8Array` to `Uint8Array` and similar give unexpected results.

ndzik (changes requested):

src/utils/arrays/index.ts
sebastianst: I really need to get used to this `===` business in JS ^^

src/utils/arrays/index.ts
sebastianst: I really need to get used to this `===` business in JS ^^

sebastianst: The nicest way would be to define an interface with property `length`, function `every` and `[]` index access. But I cannot be bothered right now to find out how this is done correctly.

ndzik (approved):
Absolute unit :mountain:

PR #45, opened 2021-08-05
Update interfaces and MockClient

sebastianst:
- Extend interfaces so that ErdstallClient includes erdstall()
- Expose erdstall() in Client impl
- Add mock initialize and subscribe to MockClient


src/client.ts
```diff
@@ -27,13 +27,15 @@ export class Client implements ErdstallClient {
...
-			this.enclaveConn = Enclave.dial(encConn! as URL);
+			this.enclaveConn = Enclave.dial(encConn as URL);
 		else
-			this.enclaveConn = encConn! as EnclaveReader;
+			this.enclaveConn = encConn as EnclaveReader;
 		this.erdstallEventHandlerCache = new EventCache<ErdstallEvent>();
 		this.erdstallOneShotHandlerCache = new OneShotEventCache<ErdstallEvent>();
 	}
 
+	erdstall = this.erdstallConn!.erdstall;
```
RmbRT, line 37: I think this executed only once, just before the constructor. A function would be better.

RmbRT (changes requested):

src/client.ts
```diff
@@ -27,13 +27,17 @@ export class Client implements ErdstallClient {
...
+			this.enclaveConn = Enclave.dial(encConn as URL);
 		else
-			this.enclaveConn = encConn! as EnclaveReader;
+			this.enclaveConn = encConn as EnclaveReader;
 		this.erdstallEventHandlerCache = new EventCache<ErdstallEvent>();
 		this.erdstallOneShotHandlerCache = new OneShotEventCache<ErdstallEvent>();
 	}
 
+	erdstall(): Address {
+		return this.erdstallConn!.erdstall()
```
RmbRT, line 38: Missing `;`

RmbRT (changes requested):
*You need to leave a comment to request changes.*

src/client.ts
sebastianst: ever executed `yarn lint` lately? ;)

RmbRT (approved):

PR #47, opened 2021-08-06
Session: Update nonce on error

ndzik:


src/session.ts
```diff
@@ -31,7 +41,11 @@ export class Session extends Client implements ErdstallSession {
 		this.nonce = 0n;
 		// When encountering any error, assume it might be a nonce mismatch. In
 		// this case, reset the nonce to an invalid value.
-		this.enclaveWriter.on("error", () => { this.nonce = 0n; });
+		this.enclaveWriter.on("error", () =>
+			this.nextNonce().then((nonce) => {
+				this.nonce = nonce;
+			}),
```
sebastianst, line 47: I think you can replace this by just
```suggestion
			await this.updateNonce();
```
(or `.then()` if our `on` callbacks don't support async functions (yet)).
Because this logic in effect calls `updateNonce`, then increases it with `nonce++` but returns the previous one, and then sets it back.

sebastianst (changes requested):

sebastianst:
Fixed byhttps://github.com/perun-network/erdstall-ext/pull/70

PR #51, opened 2021-08-09
Add newRandomString

sebastianst:


ndzik (approved):
Peak performance :mount_fuji:

PR #58, opened 2021-08-19
BalanceProof encoding and signing

sebastianst:
- api/util: Rename interface ABITaggedEncodable->Packable
- ledger/assets: Change Assets.asABI to return ErdstallToken[]
- api/responses: Add abi encoding, sign and verify to Balance{,Proof}
- Move ledger test Environment setup into test/ledger
- api/responses: Test correct Balance encoding
- api/responses: Test correct balance proof signing.


src/api/responses/balanceproof.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import {
+	Erdstall,
+	Erdstall__factory,
+} from "#erdstall/ledger/backend/contracts";
+
+import * as test from "#erdstall/test";
+import { Enviroment, setupEnv } from "#erdstall/test/ledger";
+import { equalArray } from "#erdstall/utils/arrays";
+
+describe("BalanceProofs", function () {
```
ndzik, line 19: Any reason why you used `function () {...}` definition here instead of our usual `() => {...}`.

In this case they exhibit the same behavior, but I would say we should reserve `function () {...}` definition for callbacks in our testsuite, when we act on the `this` object itself. This makes it clear on a first glance, that the callback is not pure but depends on some settings of the testsuite.

For instance, with the current callback definition, you would be able to specify the following:

```ts
it("does some funny stuff", function (done) {
    this.timeout(4000); // sets the test timeout for THIS test to 4 seconds overriding the defaults.
    //...
}
```

sebastianst: I did it because [the mocha docs recommend it as default behavior](https://mochajs.org/#arrow-functions):
> Passing arrow functions (aka “lambdas”) to Mocha is discouraged. [...] If you do not need to use Mocha’s context, lambdas should work. Be aware that using lambdas will be more painful to refactor if the need eventually arises!

So I thought, why not start now? It doesn't hurt and, if anything, will be better in the future.

ndzik: Alright, they definitely have more experience so trusting their judgement shouldn't be a bad idea ^^

ndzik (approved):
Amaazing :star: :zap:

PR #59, opened 2021-08-19
Ci yarn

### LINKED ISSUE #48: Use yarn in CI
It still uses `npm`. Can also enable the `caching` option of the node setup action.

sebastianst:


ndzik:
We should probably add a step which runs `prettier` and fails if changes are suggested.

sebastianst:
> We should probably add a step which runs `prettier` and fails if changes are suggested.

Total agreement, but I'd like to do that in another PR because otherwise, I have to prettify the whole codebase now, which I'd rather reserve for that next PR.

PR #61, opened 2021-08-19
Add Address.ensure

sebastianst:
- ledger: Add Address.ensure
- api/responses: Relax Balance ctor to accept string addresses


ndzik (approved):
A piece of :art:

PR #66, opened 2021-09-01
test: Fix and change newRandomBigInt

sebastianst:
- The second parameter is now the number of bits, not the max value.
- newRandomBigInt was called NewBigInt before. Renamed it to make sure
that all users of this function update their call. We wanted to rename
it anyways.


ndzik (approved):
I love it :heart_eyes:

PR #69, opened 2021-09-02
Improve error reporting

RmbRT:


src/enclave/provider.ts
```diff
@@ -3,7 +3,7 @@
 
 import WebSocket from "isomorphic-ws";
 
-export const UninitializedConn = new Error("uninitialized connection");
+export const UninitializedConn = "uninitialized connection";
```
ndzik, line 6: The reason we had a constant `Error` was to allow easy comparisons on the user side.

`try { ... } catch (e) { if (e === sdk.UninitializedConn) /* do something nice */ }`

What was your reason behind the change?

src/enclave/provider.ts
```diff
@@ -23,23 +23,27 @@ export class EnclaveWSProvider implements EnclaveProvider {
...
-			this.ws.close();
-		}
+		this.close();
 		this.ws = new WebSocket(this.url);
 		this.ws.onmessage = this.onmessage;
 		this.ws.onerror = this.onerror;
-		this.ws.onopen = this.onopen;
+		this.ws.onopen = (e: Event): any => {
+			this.initialized = true;
+			if(this.onopen)
```
ndzik, line 44: `Prettier` kicked in for me here. Version `2.3.2`. Can you check this on your end please?

ndzik (changes requested):

src/enclave/provider.ts
RmbRT: The whole repository doesn't have prettier, yet. For me it currently lists over 70 files as ugly. I think we should make a PR that runs it over the whole repository and then also set up the compile-time check like we do in the marketplace.

src/enclave/provider.ts
RmbRT: The error's stacktrace is lost because it's set in the constructor, not upon throwing, as far as I know. I'm not sure what the best course of action is, here. I'd recommend doing a check whether `e.message === UninitializedConn`.

src/enclave/provider.ts
ndzik: Ah, so I mixed that up with our demo then :laughing:

src/enclave/provider.ts
ndzik: I will just create an issue in which we relay this discussion. I am okay with it for now, since we do not have any dependencies, which rely on the old behavior right now.

Tracked in #78

src/enclave/provider.ts
ndzik: Ironically we have a commit "incorporating prettier" but it didn't apply any changes to the files. We will do this in another PR, agree.

ndzik (approved):
Glorious, but I think this has to be rebased.

PR #72, opened 2021-09-08
Also emit existing contract declaration files

ndzik:
When building and publishing our package, we give access to the contracts we use. Although some of those definitions are usable for the user, they are lacking type information resulting in every contract related function to be of the `any` type.

Since `tsc` does not yet support emitting already existing `d.ts` files, one is required to work around this by copying/overwriting them in the output. This PR does exactly that and also updates the erdstall-contracts dependency to the latest version.


scripts/publish_posthook.ts
```diff
@@ -11,6 +12,29 @@ if (args.length != 4) {
...
+			file: (base, stats, next) => {
+				if (!stats.name.includes(".d.ts")) return next();
+
+				const relativeSrc = [base, stats.name].join("/");
+
+				const strippedBase = base.slice(src.length + 1); // +1 => remove leading "/".
+				const relativeDest = [dest, strippedBase, stats.name].join("/");
+				copyFileSync(relativeSrc, relativeDest);
+
+				next();
```
sebastianst, line 31: not `return`? because in the first check you return it...

scripts/publish_posthook.ts
```diff
@@ -11,6 +12,29 @@ if (args.length != 4) {
 	process.exit(1);
 }
 
+// copyDeclarationFilesToDestination copies the declaration files contained in
+// `src` into `dest` with their respective subfolder structure.
+// e.g.:
+//              src/x/y/z.d.ts -> dist/x/y/z.d.ts.
+function copyDeclarationFilesToDestination(src: string, dest: string) {
```
sebastianst, line 19:
```suggestion
function copyDeclarationFiles(src: string, dest: string) {
```
IMHO `ToDestination` is noise

package.json
```diff
@@ -69,7 +69,7 @@
 		"build": "ttsc -p tsconfig.build.json",
 		"watch": "ttsc -w",
 		"lint": "eslint ./src --ext .ts",
-		"bindings": "./scripts/genbindings.sh deps/erdstall-contracts",
+		"bindings": "./scripts/genbindings.sh",
```
sebastianst, line 74: hmm, the commit message says the opposite of what's happening here, no? Also, you don't update the actual script file in this commit. So it didn't need this argument in the first place?

src/ledger/backend/contracts/hardhat.d.ts
```diff
@@ -0,0 +1,107 @@
+/* Autogenerated file. Do not edit manually. */
```
sebastianst, line 1: Why do we suddenly generate a hardhat file? Do we need/want it?

sebastianst (changes requested):

scripts/publish_posthook.ts
ndzik: I just did it like this because I wanted it to be a one-liner.

```ts
if (...) return next();
```

instead of

```ts
if (...) {
    next();
    return;
}
```

package.json
sebastianst: Ahh now I get it. Could you add a small sentence to the README on how to generate new bindings?

scripts/publish_posthook.ts
sebastianst: nifty 😎

src/ledger/backend/contracts/hardhat.d.ts
ndzik: AFAIK this is for better hardhat integration and contains lots of helper types for writing tests. Might come in handy later down the road, but I have no problem with removing it.

sebastianst: I'm happy to keep it :)

sebastianst (approved):
Piece of Art 🎨

PR #75, opened 2021-09-14
test: Extend MockWatcher

### LINKED ISSUE #67: MockWatcher: allow subscribing for balance, exit proofs, phase shifts.

RmbRT:
Closes #67.

### DIFF
--- a/src/test/mocks.ts
+++ b/src/test/mocks.ts
@@ -60,13 +60,13 @@ export class MockWatcher implements Watcher {
 	}
 
 	phaseshift(bps: BalanceProofs) {
-		this.phaseShiftHandler();
 		for(let bp of bps.map.values()) {
 			if(bp.balance.exit)
 				this.exitProofHandler(bp);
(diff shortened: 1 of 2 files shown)


src/test/mocks.ts
```diff
@@ -44,6 +58,16 @@ export class MockWatcher implements Watcher {
 	transfer(tx: Transfer): void {
 		this.txReceiptHandler(new TxReceipt(tx, new Account(0n, new Assets())));
 	}
+
+	phaseshift(bps: BalanceProofs) {
+		this.phaseShiftHandler();
```
sebastianst, line 63: I think the `phaseShiftHandler` should be called after the proof handlers. Because strictly speaking, the proofs still belong to the old epoch and the phase shift is the last event before the new epoch starts.

sebastianst (changes requested):

sebastianst (approved):

PR #77, opened 2021-09-17
Burn tx

### LINKED ISSUE #74: Expose Burn Transaction
Will be introduced in https://github.com/perun-network/erdstall-ext/issues/84

RmbRT:
Closes #74.

### DIFF
--- a/src/e2e/sdk_remote_integration.spec.ts
+++ b/src/e2e/sdk_remote_integration.spec.ts
@@ -133,4 +133,14 @@ describe("Erdstall-TS-SDK", () => {
 		const tradeOffer = await clients[2].createOffer(offer, expect);
 		return clients[3].acceptTrade(tradeOffer);
 	});
+
+	it("allows burning tokens", async() => {
+		const token = clients[2].address;
+		const rec = await clients[3].burn(new Assets({
+			token,
+			asset: new assets.Tokens([nft]),
+		}));
+		if(rec.account.values.hasAsset(token))
+			throw new Error("expected asset to be burnt");
+	});
 });
--- a/src/e2e/sdk_remote_integration.spec.ts
+++ b/src/e2e/sdk_remote_integration.spec.ts
@@ -80,7 +80,18 @@ describe("Erdstall-TS-SDK", () => {
 		});
 
 		const stages = await Promise.all(clients.map(c => c.deposit(depositBal)));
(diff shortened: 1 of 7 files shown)


src/e2e/sdk_remote_integration.spec.ts
```diff
@@ -83,6 +83,16 @@ describe("Erdstall-TS-SDK", () => {
 		return Promise.all(stages.map(stage => stage.wait()));
 	});
 
+	it("await deposit finalization", async() => {
```
ndzik, line 86: Is this a testcase? I would prefer to have it in a normal closure which is used to wait for phaseshifts, otherwise every `it` is potentially not a testcase.

ndzik: I just realise that quite a few of those testcases are no testcases but rather steps to setup the environment, or am I missing something?

ndzik: I realise this is not the scope of this PR, so this should not be done here but IMHO I think we should extract those steps into a dedicated `setup` function which then also returns a helper object with which one can wait for phaseshifts. These could then be utilised with the `before{All|Each}` hooks.

If you agree, I would appreciate an issue for this :smile:

ndzik (changes requested):
The changes incorporating the burn tx itself are flawless, except for the fact that a `newRandomBurn` is missing in `test/transactions/transactions.ts`.

src/e2e/sdk_remote_integration.spec.ts
RmbRT: 1. `it()` is asynchronous, and I didn't know another way of executing stuff between two tests. For now, I included it in the deposit test.
2. Stuff like subscribing clients could be seen as a test case, as subscribing isn't tested anywhere else against a real operator, AFAIK, but it's also necessary for the next tests.
3. I think I'll have to take another look at how mocha works.

ndzik: We will keep it like this for now, I created an [issue](https://github.com/perun-network/erdstall-ts-sdk/issues/79) in which this can be investigated further. As far as I am aware, it is common to create sub-testsuites (with `describe`) in which one can run the needed code in between multiple test(-suites).

ndzik (approved):
Like a butterfly :butterfly:

PR #81, opened 2021-09-21
Add metadata provider

ndzik:
Adds definitions for the NFTMetadataProvider.

Currently blocked.


ndzik:
Two mentionable points:

1. I think it would be nice to create a solidity contract defining an "umbrella" interface for metadata stuff. We are currently abusing the fact, that the metadata of all contracts for which metadata is requested, implements the `ERC721IMetadata` interface. Any opinions?

2. I fixed a typo in our naming and I am certain it will break stuff in other repositories.

src/ledger/backend/readconn.ts
```diff
@@ -48,4 +56,25 @@ export class LedgerReadConn implements LedgerReader {
...
+
+	async getNftMetadata(token: Address, id: bigint): Promise<NFTMetadata> {
+		const tokenS = token.toString();
+		if (this.metadataUriCache.has(tokenS)) {
+			return this.metadataUriCache.get(tokenS)!;
+		}
+
+		// TODO: Create IMetadata contract interface for which bindings can be
+		// generated, so we have a single interface over which we can query
+		// metadata for different contract implementations in the future.
```
sebastianst, line 68: What other token standards, besides ERC721, have the exact same public function `tokenURI(uint256)`?

src/ledger/backend/readconn.ts
```diff
@@ -12,15 +15,20 @@ export const ErrErdstallContractNotConnected = new Error(
...
 
 export class LedgerReadConn implements LedgerReader {
 	readonly contract: Erdstall;
 	private eventCache: Map<Function, (args: Array<any>) => void>;
+	private metadataUriCache: Map<string, NFTMetadata>;
 
 	constructor(contract: Erdstall) {
 		this.contract = contract;
 		this.eventCache = new Map<Function, (args: Array<any>) => void>();
+		this.metadataUriCache = new Map<string, NFTMetadata>();
```
sebastianst, line 31: you can set those two directly where you define the fields, then you don't need to set them in the ctor.

src/ledger/backend/readconn.ts
```diff
@@ -12,15 +15,20 @@ export const ErrErdstallContractNotConnected = new Error(
...
-export interface LedgerReader extends ErdstallWatcher, Contracter {}
+export interface LedgerReader
+	extends NFTMetadataProvider,
+		ErdstallWatcher,
+		Contracter {}
 
 export class LedgerReadConn implements LedgerReader {
 	readonly contract: Erdstall;
 	private eventCache: Map<Function, (args: Array<any>) => void>;
+	private metadataUriCache: Map<string, NFTMetadata>;
```
sebastianst, line 26: It really is a cache for metadata itself, not for the URIs. So either rename it to `metadataCache` or if your intention was to only cache URIs, then you need to change the logic in `getNftMetadata` and change the Map value type to strings/URIs.

src/ledger/backend/readconn.ts
```diff
@@ -48,4 +56,25 @@ export class LedgerReadConn implements LedgerReader {
 	erdstall(): Address {
 		return Address.fromString(this.contract.address);
 	}
+
+	async getNftMetadata(token: Address, id: bigint): Promise<NFTMetadata> {
```
sebastianst, line 60: If we keep metadata caching, we should give the user a way to make a fresh request, e.g. by adding to this function (and interface) an additional parameter `useCache?: bool`. I'm happy with keeping the default to true.

src/client.ts
```diff
@@ -41,6 +41,13 @@ export class Client implements ErdstallClient {
 		return this.erdstallConn!.erdstall();
 	}
 
+	getNftMetadata(token: Address, id: bigint): Promise<NFTMetadata> {
```
sebastianst, line 45: also add `useCache?` bool here

src/test/mocks.ts
```diff
@@ -91,6 +92,9 @@ export class MockClient extends MockWatcher implements ErdstallClient {
 	async getAccount(_who: Address): Promise<Account> {
 		throw new Error("cannot query accounts on mock clients");
 	}
+	async getNftMetadata(_token: Address, _id: bigint): Promise<NFTMetadata> {
+		throw new Error("cannot query nft data on mock clients");
+	}
```
sebastianst, line 113: When testing the marketplace server, we might need to inject metadata. So I'd be great if you could put a `metadata: Map<string, NFTMetadata>` into the mock client with a `setMetadata(token, id, NFTMetadata)` setter whose values are then read here.

src/ledger/backend/connection.spec.ts
```diff
@@ -52,4 +55,24 @@ describe("ErdstallConnection", () => {
...
+		);
+		const bp = await bal.sign(conn.erdstall(), testenv.tee);
+		const stages = await conn.withdraw(bp);
+		for (const stage of stages) {
+			const ctx = await stage.value;
+			const rec = await ctx.wait();
+			expect(rec.status, "withdrawing should have worked").to.equal(0x1);
+		}
+
+		return withdrawRegistered;
```
sebastianst, line 76: What's mocha gonna do with this return value? Note that I don't understand/know the `EventHelper` ;)

src/test/metadata.ts
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: Apache-2.0
+"use strict";
+
+import PRNG, { newRandomUint8Array } from "./random";
```
sebastianst, line 4: Note that we also have `newRandomString` which might be more appropriate here

sebastianst (changes requested):

src/ledger/backend/readconn.ts
ndzik: Any advantage you can point out? The way I always view it is that the member definitions (unitialised) describe the shape while the constructor always initializes (paints the picture :art:).

One could even go further and completely skip the member definition for passed params and define and default assign them in the constructor arguments. There is so many ways and I **think** we currently mainly first define and then assign them separately.

src/ledger/backend/readconn.ts
sebastianst: Let's think about generalization when we start handling other token types. E.g., in the case of 1155, has it the same interface with same parameters?

src/ledger/backend/readconn.ts
ndzik: Good point!

src/ledger/backend/readconn.ts
ndzik: Most likely not. I just thought of some exotic Tokens which might be used in the wild but differ in the way they report their `tokenURI()`. But you are right, let's talk about it when it really matters again ^^

src/ledger/backend/connection.spec.ts
ndzik: Mocha just waits for the `Promise<boolean>` to resolves and asserts that it is `true` and did not fail otherwise.

`withdrawRegistered` here is just a sanity check, that the `Withdraw` event was really triggered, with mocha waiting a maximum of 10 seconds for it.

src/ledger/backend/readconn.ts
ndzik: But I can shorten the constructors, because the types can be inferred. Quite old code right here.

src/ledger/backend/metadata.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+		// Second request should be cached, resulting in `nock` being unable to
+		// intercept a GET request.
+		nock(PERUNART_URI)
+			.get(`/${token.toString().toLowerCase()}/${id}`)
+			.replyWithError("this should never be called");
+		res = await erdstall.getNftMetadata(token, id);
+		expect(res).to.deep.equal(metadata);
+
+		// Clean up interception.
+		nock.cleanAll();
```
sebastianst, line 57: so this isn't needed at the end again? E.g. if the client doesn't call it properly again, the next nock interception would exist outside this test?

src/ledger/backend/metadata.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+		// Third request with a forced fetch should query the `tokenURI` again.
+		nock(PERUNART_URI)
+			.defaultReplyHeaders({ "access-control-allow-origin": "*" })
+			.get(`/${token.toString().toLowerCase()}/${id}`)
+			.reply(200, {});
+		res = await erdstall.getNftMetadata(token, id, false);
+		expect(res).to.not.deep.equal(metadata);
```
sebastianst, line 65: Couldn't you be even more specific here and say `.to.deep.equal({})` or even `to.be.empty`?

src/ledger/backend/metadata.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+		expect(res).to.deep.equal(metadata);
+
+		// Clean up interception.
+		nock.cleanAll();
+
+		// Third request with a forced fetch should query the `tokenURI` again.
+		nock(PERUNART_URI)
+			.defaultReplyHeaders({ "access-control-allow-origin": "*" })
+			.get(`/${token.toString().toLowerCase()}/${id}`)
+			.reply(200, {});
```
sebastianst, line 63: You could set some field
```suggestion
			.reply(200, {image: "foo"});
```
and then assert the field value. More explicit than just asserting `to.not.deep.equal`.

src/test/mocks.ts
```diff
@@ -92,7 +111,132 @@ export class MockClient extends MockWatcher implements ErdstallClient {
 		throw new Error("cannot query accounts on mock clients");
 	}
 
+	setMetadata(token: Address, id: bigint, metadata: NFTMetadata): void {
+		this.metadata.set(`${token}:${id}`, metadata);
+	}
+
+	async getNftMetadata(token: Address, id: bigint): Promise<NFTMetadata> {
```
sebastianst, line 118: Don't you miss `useCache?: boolean` here now? Otherwise the mock doesn't implement the interface.

src/test/mocks.ts
```diff
@@ -92,7 +111,132 @@ export class MockClient extends MockWatcher implements ErdstallClient {
...
+		return this.metadata.get(`${token}:${id}`)!;
```
sebastianst, line 119: Maybe we should return a rejecting promise if the value is undefined? This is how axios would behave, right?

sebastianst (changes requested):

src/ledger/backend/metadata.spec.ts
ndzik: Added a `afterEach()` hook, which handles this case. Again, good argument.

src/test/mocks.ts
ndzik: TBH I was wondering the same initially. But truth is defining an `interface` describing some `function` with optional parameters, also makes them optional for the implementer. I expect it to work, because JS happily drops superfluous arguments.

sebastianst: I see, we're in JS land, where this makes sense. If the interface check succeeds, it should be fine.

sebastianst (approved):
fantastic ✨

PR #82, opened 2021-09-22
Make PerunArt contract available in Testenvironment and test Deposits

ndzik:
Includes changes to the .gitignore for autogenerated files by hardhat and typechain, as well as a simple implementation of an ERC721 contract which allows everyone to mint, making it possible in tests to have external NFTs.
These external NFTs also have a `tokenURI` set, which can be mocked in tests using stuff like `nock`.


ndzik:
Needed for #81

scripts/genbindings.sh
```diff
@@ -28,6 +28,24 @@ function checkInstalledLocally() {
...
+    echo compileAndMoveTo [source] [destination]
+    exit 1
+  fi
+
+  CWD="${PWD}"
+  cd "${1}"
+
+  mkdir -p "${2}"/contracts
+  mkdir -p "${2}"/contracts/abi
+  npx hardhat compile && cp -r ./typechain/* "${2}"/contracts/ \
```
sebastianst, line 43: We use `yarn` everywhere, so I propose
```suggestion
  yarn hardhat compile && cp -r ./typechain/* "${2}"/contracts/ \
```

src/test/ledger/env.ts
```diff
@@ -24,11 +28,16 @@ export interface Enviroment {
...
 	erc721Holder: string;
+	perunArt: string;
 	op: Wallet;
 	tee: Wallet;
 	users: Wallet[];
 }
 
+const PERUNART_NAME = "PerunArt";
+const PERUNART_SYMBOL = "PART";
+const PERUNART_URI = "https://nifty.erdstall.dev";
```
sebastianst, line 39: always needs trailing `/` because it's concatenated using `abi.encodePacked(uriBase, tokenId)`
```suggestion
const PERUNART_URI = "https://nifty.erdstall.dev/";
```

src/ledger/backend/contracts_deposit.spec.ts
```diff
@@ -4,38 +4,58 @@
...
+	const amount = new Amount(test.newRandomBigInt(rng, 16));
+	const tokens = test.newRandomTokens(rng, TOKEN_SIZE);
 
 	before(async () => {
 		testenv = await setupEnv();
 		bob = testenv.users[0];
+
+		const part = PerunArt__factory.connect(testenv.perunArt, bob);
+		for (const id of tokens.value) {
+			await part.mint(bob.address, id);
```
sebastianst, line 35: So `bob` is configured as being a minter before? Otherwise this `mint` call will revert.

scripts/genbindings.sh
```diff
@@ -28,6 +28,24 @@ function checkInstalledLocally() {
...
+  if [ ! -d "$1" ] || [ ! -d "$2" ]; then
+    echo ERROR: No valid source or destination specified
+    echo compileAndMoveTo [source] [destination]
+    exit 1
+  fi
+
+  CWD="${PWD}"
+  cd "${1}"
+
+  mkdir -p "${2}"/contracts
```
sebastianst, line 41: you can delete this line, since you use `-p` in the next

scripts/genbindings.sh
sebastianst: Why copy it instead of moving it?
```suggestion
  npx hardhat compile && mv ./typechain/* "${2}"/contracts/ \
```

scripts/genbindings.sh
```diff
@@ -28,6 +28,24 @@ function checkInstalledLocally() {
...
+  mkdir -p "${2}"/contracts/abi
+  npx hardhat compile && cp -r ./typechain/* "${2}"/contracts/ \
+    && find ./artifacts/contracts/*.sol/ \
+    | grep -v ".dbg" | grep ".json" | xargs cp -t "${2}"/contracts/abi/
```
sebastianst, line 45: again, can we move?

scripts/genbindings.sh
```diff
@@ -28,6 +28,24 @@ function checkInstalledLocally() {
...
+  cd "${CWD}"
```
sebastianst, line 46: is there any cleanup of generated intermediary files happening somewhere?

scripts/genbindings.sh
```diff
@@ -28,6 +28,24 @@ function checkInstalledLocally() {
   fi
 }
 
+function compileAndMoveTo() {
```
sebastianst, line 31: That the files are moved are an implementation detail of this function IMO, so I'd just call it 
```suggestion
function compileTo() {
```
;) Also, right now they're copied.

src/ledger/backend/contracts_deposit.spec.ts
```diff
@@ -4,38 +4,58 @@
...
+		const testCases: [DepositerCallsFactory, string, string, Asset][] = [
+			[makeETHDepositCalls, testenv.ethHolder, ETHZERO, amount],
+			[makeERC20DepositCalls, testenv.erc20Holder, testenv.perun, amount],
+			[
+				makeERC721DepositCalls,
+				testenv.erc721Holder,
+				testenv.perunArt,
+				tokens,
+			],
+		];
```
sebastianst, line 51: So it's never tested that the `Session` itself implements `deposit` correctly, but only the intermediary functions `makeFooDepositCalls`?

sebastianst (changes requested):

src/test/ledger/env.ts
ndzik: :eagle: :eye:

src/ledger/backend/contracts_deposit.spec.ts
ndzik: Very good. I forgot extending the `LedgerWriter` spec to also deposit `Tokens` :rocket: Added.

ndzik: Which itself is just exposed by the `Session` implementation.

scripts/genbindings.sh
ndzik: Now `cp -rl`'ing the required generated bindings/abis. Forced us to remove existing bindings beforehand, which had the nice side effect of removing out of date openzeppelin contracts.

src/ledger/backend/contracts_deposit.spec.ts
ndzik: Yes, in the end to end test. But it is still missing a erc721 deposit case.

ndzik: The only reason I did not bother adding an extra testcase there, although beneficial in the long run, is because I could not immediately wrap my head around how the end to end test is structured.
And considering the `Session.deposit(...)` function looks like this:

```ts
	async deposit(
		assets: Assets,
	): Promise<Stages<Promise<ethers.ContractTransaction>>> {
		if (!this.erdstallConn) {
			return Promise.reject(ErrUnitialisedClient);
		}

		return (this.erdstallConn as LedgerWriter).deposit(assets);
	}
```

Which is essentially just `this.erdstallConn.deposit(assets)`. And this function is tested already.

This is just the quickshot, I can dig into the end to end test and setup another testcase there.

sebastianst: OK thanks, I guess it is fine for now then. Would still be good to add an issue about it though, so we don't forget...

ndzik: Tracked in #86

sebastianst (approved):

PR #84, opened 2021-09-24
asset: Fix and test Tokens.cmp()

RmbRT:
Needed for deposit logic in nerd marketplace.

### DIFF
--- a/src/ledger/assets/tokens.spec.ts
+++ b/src/ledger/assets/tokens.spec.ts
@@ -31,4 +31,38 @@ describe("Tokens", () => {
 			t1.sub(t2);
 		}, "subbing incompatible tokens should throw").throws;
 	});
+
+	it("compares tokens", () => {
+		expect(new Tokens([124n]).cmp(new Tokens([]))).to.eql("gt");
+		expect(new Tokens([124n]).cmp(new Tokens([124n]))).to.eql("eq");
+		expect(new Tokens([125n]).cmp(new Tokens([124n]))).to.eql(
+			"uncomparable",
+		);
+		expect(new Tokens([125n]).cmp(new Tokens([124n, 125n]))).to.eql("lt");
+		expect(new Tokens([126n]).cmp(new Tokens([124n, 125n]))).to.eql(
+			"uncomparable",
+		);
+		expect(new Tokens([126n]).cmp(new Tokens([124n, 125n]))).to.eql(
+			"uncomparable",
+		);
+		expect(
+			new Tokens([123n, 125n]).cmp(new Tokens([123n, 124n, 125n, 126n])),
+		).to.eql("lt");
+		expect(
+			new Tokens([123n, 124n]).cmp(new Tokens([122n, 123n, 124n, 125n])),
+		).to.eql("lt");
+		expect(
+			new Tokens([123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
+		).to.eql("gt");
+		expect(
+			new Tokens([122n, 123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
+		).to.eql("gt");
+	});
+
+	it("sorts tokens", () => {
+		expect(() => new Tokens([123n,124n, 125n, 124n])).throws;
+		expect(() => new Tokens([1n, 2n, -3n])).throws;
+		expect(new Tokens([1234n, 123n, 234n]).value).to.eql([123n, 234n, 1234n]);
+		expect(new Tokens([1n, 10n, 2n]).value).to.eql([1n, 2n, 10n]);
+	});
 });
--- a/src/ledger/assets/tokens.ts
+++ b/src/ledger/assets/tokens.ts
@@ -20,7 +20,18 @@ export class Tokens extends Asset {
 
 	constructor(v: bigint[]) {
 		super();
-		this.value = v;
+		this.value = v.sort((a, b) => (a < b ? -1 : a == b ? 0 : 1));
+		Tokens.requireUniquePositiveEntries(this.value);
+	}
+
+	// requireUniquePositiveEntries expects `v` to be sorted and asserts that every
+	// entry in `v` is unique and a positive `bigint`. If the invariant does not
+	// hold, an error is thrown.
+	private static requireUniquePositiveEntries(v: bigint[]) {
+		for (let k = 0; k < v.length - 1; k++) {
+			if (v[k] < 0n) throw new Error("invalid token id in token set");
+			if (v[k] === v[k + 1]) throw new Error("token entries not unique");
+		}
 	}
 
 	toJSON() {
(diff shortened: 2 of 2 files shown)


src/ledger/assets/tokens.ts
```diff
@@ -65,36 +65,28 @@ export class Tokens extends Asset {
 			return "uncomparable";
 		}
 		let res = ["lt", "eq", "gt"] as const;
-		let a = this.value;
-		let b = (asset as Tokens).value;
+		let a = this.value.sort();
```
ndzik, line 68: IF this is really required we have another problem with the methods `addID` and `sub`, because these are the only functions with which the `Tokens` should be populated. Also the constructor is missing a check that the given `bigint[]` array is already sorted (or sorted when constructing).

ndzik: In short, `this.value.sort()` should never be required.

src/ledger/assets/tokens.ts
```diff
@@ -65,36 +65,28 @@ export class Tokens extends Asset {
...
+		let b = (asset as Tokens).value.sort();
```
ndzik, line 69:
```suggestion
		let a = this.value;
		let b = (asset as Tokens).value;
```

ndzik: And add a `requireUniquePositiveEntries` function which is used in the constructor after sorting the input array, something like:

```ts
// requireUniquePositiveEntries expects `v` to be sorted and asserts that every
// entry in `v` is unique and a positive `bigint`. If the invariant does not
// hold, an error is thrown.
function requireUniquePositiveEntries(v: bigint[]) {
	for (let k = 0; k < v.length - 1; k++) {
		if (v[k] < 0n) throw new Error("invalid token id in token set");
		if (v[k] === v[k + 1]) throw new Error("no unique token entries");
	}
}
```

src/ledger/assets/tokens.spec.ts
```diff
@@ -31,4 +31,12 @@ describe("Tokens", () => {
 			t1.sub(t2);
 		}, "subbing incompatible tokens should throw").throws;
 	});
+
+	it("compares tokens", () => {
+		expect(new Tokens([124n]).cmp(new Tokens([]))).to.eql("gt");
+		expect(new Tokens([124n]).cmp(new Tokens([124n]))).to.eql("eq");
+		expect(new Tokens([125n]).cmp(new Tokens([124n]))).to.eql("uncomparable");
+		expect(new Tokens([125n]).cmp(new Tokens([124n, 125n]))).to.eql("lt");
+		expect(new Tokens([126n]).cmp(new Tokens([124n, 125n]))).to.eql("uncomparable");
```
ndzik, line 40:
```suggestion
		expect(new Tokens([126n]).cmp(new Tokens([124n, 125n]))).to.eql("uncomparable");
				expect(
			new Tokens([123n, 125n]).cmp(new Tokens([123n, 124n, 125n, 126n])),
		).to.eql("lt");
		expect(
			new Tokens([123n, 124n]).cmp(new Tokens([122n, 123n, 124n, 125n])),
		).to.eql("lt");
		expect(
			new Tokens([123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
		).to.eql("gt");
		expect(
			new Tokens([122n, 123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
		).to.eql("gt");
```

Just more edge cases to be on the safe side, because  I saw these ones failing when double checking our go-impl...

ndzik: And you apparently do not run prettier? We have a `.prettierrc` but it kicked in again.

ndzik (changes requested):

src/ledger/assets/tokens.ts
```diff
@@ -65,36 +65,28 @@ export class Tokens extends Asset {
...
-			if (lhs < rhs) {
-				cmp = 1;
-			} else if (lhs === rhs) {
-				bIdx++;
-			} else if (lhs > rhs) {
-				break;
-			}
+		for (let i = 0; i < a.length && bIdx < b.length; i++) {
+			if (a[i] == b[bIdx])
+				++bIdx;
```
ndzik, line 82:
```suggestion
			if (a[i] == b[bIdx]) ++bIdx;
			else if (a[i] > b[bIdx]) break; // short circuit.
```

ndzik (approved):
Splendid.

src/ledger/assets/tokens.ts
```diff
@@ -20,8 +20,18 @@ export class Tokens extends Asset {
 
 	constructor(v: bigint[]) {
 		super();
+		Tokens.requireUniquePositiveEntries(v);
 		this.value = v;
 	}
```
sebastianst, line 25: If our data model _requires_ `v` to be sorted, because the class's implementations don't work otherwise, shouldn't we sort it in the constructor? I'd create a sorted copy of `v` to `this.value` and then call `requireUniquePositiveEntries` on it.

sebastianst (changes requested):
one open question

src/ledger/assets/tokens.ts
ndzik: Agree.

RmbRT: ok!

src/ledger/assets/tokens.ts
```diff
@@ -20,7 +20,18 @@ export class Tokens extends Asset {
 
 	constructor(v: bigint[]) {
 		super();
-		this.value = v;
+		Tokens.requireUniquePositiveEntries(v);
+		this.value = v.sort((a, b) => (a < b ? -1 : a == b ? 0 : 1));
```
ndzik, line 24:
```suggestion
              this.value = v.sort((a, b) => (a < b ? -1 : a == b ? 0 : 1));
		Tokens.requireUniquePositiveEntries(this.value);
```

ndzik (changes requested):

src/ledger/assets/tokens.ts
sebastianst: agree, also what I suggested

src/ledger/assets/tokens.spec.ts
```diff
@@ -31,4 +31,31 @@ describe("Tokens", () => {
...
+		).to.eql("lt");
+		expect(
+			new Tokens([123n, 124n]).cmp(new Tokens([122n, 123n, 124n, 125n])),
+		).to.eql("lt");
+		expect(
+			new Tokens([123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
+		).to.eql("gt");
+		expect(
+			new Tokens([122n, 123n, 124n, 125n]).cmp(new Tokens([123n, 124n])),
+		).to.eql("gt");
```
sebastianst, line 59: Is sorting tested anywhere? otherwise, please add a `Tokens` constructor test that checks that the tokens are created in the right sorted order for some shuffled and also reverse sorted input array. also test that the ctor throws if an element is present twice.

sebastianst (changes requested):

sebastianst (approved):
Awesome 🚀

ndzik (approved):
I had a strange bug where Github would not allow me to comment and approve this PR.

Anyway, could not get any better :rocket:

PR #89, opened 2021-09-29
Lift `TokenTypesCache` into `ErdstallClient`

ndzik:
Adds an abstraction for the `TokenTypesCache` called `TokenCache`.
This `TokenCache` is available in `Ledger{Reader|Writer}` and can be externally owned, like the `ErdstallClient` does now.

### DIFF
--- a/src/erdstall.ts
+++ b/src/erdstall.ts
@@ -10,6 +10,7 @@ import { TxReceipt } from "#erdstall/api/responses";
 import { TradeOffer } from "#erdstall/api/transactions";
 import { BalanceProof } from "#erdstall/api/responses";
 import { Address, Account, ErdstallEvent } from "#erdstall/ledger";
+import { TokenCache } from "#erdstall/ledger/backend";
 import { Assets } from "#erdstall/ledger/assets";
 import { Uint256 } from "#erdstall/api/util";
 import { Stages } from "#erdstall/utils";
@@ -105,7 +106,9 @@ export interface ErdstallClient
 		Initializer,
 		Subscriber,
 		NFTMetadataProvider,
-		AccountGetter {}
+		AccountGetter {
+	readonly tokenCache: TokenCache;
+}
 
 export interface ErdstallSession
 	extends ErdstallClient,
--- a/src/erdstall.ts
+++ b/src/erdstall.ts
@@ -10,7 +10,7 @@ import { TxReceipt } from "#erdstall/api/responses";
 import { TradeOffer } from "#erdstall/api/transactions";
 import { BalanceProof } from "#erdstall/api/responses";
 import { Address, Account, ErdstallEvent } from "#erdstall/ledger";
-import { TokenCache } from "#erdstall/ledger/backend";
+import { TokenProvider } from "#erdstall/ledger/backend";
 import { Assets } from "#erdstall/ledger/assets";
 import { Uint256 } from "#erdstall/api/util";
 import { Stages } from "#erdstall/utils";
@@ -107,7 +107,7 @@ export interface ErdstallClient
 		Subscriber,
 		NFTMetadataProvider,
 		AccountGetter {
-	readonly tokenCache: TokenCache;
+	readonly tokenProvider: TokenProvider;
 }
 
 export interface ErdstallSession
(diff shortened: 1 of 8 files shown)


src/ledger/backend/tokencache.ts
```diff
@@ -20,10 +20,31 @@ interface Responder {
 	symbol(): Promise<string>;
 }
 
+export interface TokenCache {
+	readonly typeCache: TokenTypes;
+	readonly holderCache: TokenHolders;
+	setType(tokenAddr: string, ttype: TokenType): void;
```
ndzik, line 26: This might be redundant, but I could not decide.

sebastianst: I think it makes sense to have a `setType` method. Or what do you mean?

src/ledger/backend/tokencache.ts
```diff
@@ -20,10 +20,31 @@ interface Responder {
 	symbol(): Promise<string>;
 }
 
+export interface TokenCache {
+	readonly typeCache: TokenTypes;
+	readonly holderCache: TokenHolders;
```
sebastianst, line 25: Why do we expose them in the interface? Isn't this an implementation detail?

src/ledger/backend/tokencache.ts
```diff
@@ -20,10 +20,31 @@ interface Responder {
...
+		erdstall: Erdstall,
+		fromBlock?: number,
+	): Promise<TokenRegisteredEvent[]>;
+}
+
 type TokenTypes = Map<string, TokenType>;
 type TokenHolders = Map<TokenType, string>;
 
-export class TokenTypesCache {
+export class TokenTypesCache implements TokenCache {
```
sebastianst, line 47: Hmm, I find the naming right now somewhat confusing. Maybe rename the interface to something like `TokenProvider` and the implementation to something like `TokenFetcher`?

src/test/mocks.ts
```diff
@@ -96,13 +100,15 @@ export class MockWatcher implements Watcher {
...
 export class MockClient extends MockWatcher implements ErdstallClient {
+	readonly tokenCache: TokenCache;
 	private readonly contract: Address;
 	private metadata: Map<string, NFTMetadata>;
 
 	constructor(contract: Address) {
 		super();
 		this.contract = contract;
 		this.metadata = new Map();
+		this.tokenCache = new TokenTypesCache();
```
sebastianst, line 111: I think for the mock we actually need a `MockTokenProvider` that never queries anything and all types and holders have to be set with setters. But let's postpone the implementation of this when it's needed :)

src/erdstall.ts
```diff
@@ -105,7 +106,9 @@ export interface ErdstallClient
 		Initializer,
 		Subscriber,
 		NFTMetadataProvider,
-		AccountGetter {}
+		AccountGetter {
+	readonly tokenCache: TokenCache;
```
sebastianst, line 110: If we rename `TokenCache` to `TokenProvider`, I'd also rename this then to `tokenProvider`.

sebastianst (changes requested):

src/ledger/backend/tokencache.ts
ndzik: It is useful, I just should have removed the `typeCache` and `holderCache` from the interface and this question would not have come up :clown_face:

src/ledger/backend/tokencache.ts
ndzik: I left them in initially because I was refactoring. The next step should have been to get rid of them and fix any upcoming errors, but I seemingly forgot about that and instead decided a leaky abstraction is better :laughing: 

Fixed :rocket:

src/ledger/backend/tokencache.ts
ndzik: Naming is hard and I really prefer your suggestion.

src/test/mocks.ts
ndzik: Agree, I did not want to create a `MockTokenProvider`, because I also deemed it not to be necessary currently. I will create an issue for this though.

sebastianst (approved):
stellar 🌟

PR #94, opened 2021-10-06
ledger/assets: Add forEachNFT

sebastianst:


ndzik (approved):
I also want this function for Go (:<

PR #102, opened 2021-10-26
Add onChainQuery member to Client, needed for #152 in NERD

Ruben-Kruepper:
This PR introduces a new `OnChainQuery` interface which is satisfied by the new member `onChainQuery` added to the `ErdstallSessoin`. This member should group (future) on-chain functions which are not necessarily associated with Erdstall.

### DIFF
--- a/src/ledger/backend/ethereumOnChainQuerier.spec.ts
+++ b/src/ledger/backend/ethereumOnChainQuerier.spec.ts
@@ -1,0 +1,148 @@
+"use strict";
+
+import { expect } from "chai";
+import { EthereumOnChainQuerier } from "./ethereumOnChainQuerier";
+import { OnChainQuerier } from "../onChainQuerier";
+
+import * as test from "#erdstall/test";
+import { Environment, setupEnv } from "test/ledger";
+import { Wallet } from "@ethersproject/wallet";
+import { IERC721Minter__factory, IERC721__factory } from ".";
+import { Tokens } from "../assets";
+import PRNG from "#erdstall/test/random";
+import { afterEach } from "mocha";
+
+const PRIMES = [2, 3, 5, 7];
+const TOKEN_SIZE = PRIMES.length;
+const RNG: PRNG = test.newPrng();
+
+interface Entity {
+	wallet: Wallet;
+	tokens: Tokens;
+	runningTokens: bigint[];
+}
+
+describe("EthereumOnChainQuerier", function () {
+	let testenv: Environment;
+	let alice: Entity;
+	let bob: Entity;
+	let onChainQueryA: OnChainQuerier;
+	let aTransferSeed: number;
+	let bTransferSeed: number;
+
+	const initEntity = async (wallet: Wallet): Promise<Entity> => {
+		const tokens = test.newRandomTokens(RNG, TOKEN_SIZE);
+		const entity: Entity = {
+			wallet,
+			tokens,
+			runningTokens: Array.from(tokens.value),
+		};
+
+		const erc721Minter = IERC721Minter__factory.connect(
+			testenv.perunArt,
+			entity.wallet,
+		);
+		let nonce = await entity.wallet.provider.getTransactionCount(
+			entity.wallet.address,
+		);
+		for (const id of entity.tokens.value) {
+			await erc721Minter.mint(entity.wallet.address, id, {
+				nonce: nonce++,
+			});
+		}
+		await testenv.mine();
+
+		return entity;
+	};
+
+	const seededTokenTransfer = async (
+		from: Entity,
+		to: Entity,
+		seed: number,
+		tokenPool: Tokens,
+		mineEachTX: boolean,
+	): Promise<void> => {
+		const ierc721 = IERC721__factory.connect(testenv.perunArt, from.wallet);
+		let nonce = await from.wallet.provider.getTransactionCount(
+			from.wallet.address,
+		);
+		for (let index = 0; index < TOKEN_SIZE; ++index) {
+			if ((seed % PRIMES[index]) % 2 === 0) {
+				const token = tokenPool.value[index];
+				await ierc721.transferFrom(
+					from.wallet.address,
+					to.wallet.address,
+					token,
+					{ nonce: nonce++ },
+				);
+				to.runningTokens.push(token);
+				from.runningTokens.splice(from.runningTokens.indexOf(token), 1);
+				if (mineEachTX) {
+					await testenv.mine();
+				}
+			}
+		}
+		await testenv.mine();
+	};
+
+	before(async function () {
+		testenv = await setupEnv(2, undefined, { blockTime: 1 });
+		alice = await initEntity(testenv.users[0]);
+		bob = await initEntity(testenv.users[1]);
+		onChainQueryA = new EthereumOnChainQuerier(alice.wallet);
+	});
+
+	const transferTest = async (mineEachTX: boolean): Promise<void> => {
+		aTransferSeed = RNG.uInt32();
+		await seededTokenTransfer(
+			alice,
+			bob,
+			aTransferSeed,
+			alice.tokens,
+			mineEachTX,
+		);
+		bTransferSeed = RNG.uInt32();
+		await seededTokenTransfer(
+			bob,
+			alice,
+			bTransferSeed,
+			bob.tokens,
+			mineEachTX,
+		);
+
+		let aliceTokens = await onChainQueryA.queryTokensOwnedByAddress(
+			testenv.perunArt,
+			alice.wallet.address,
+		);
+		expect(aliceTokens.value)
+			.to.have.members(alice.runningTokens)
+			.and.length(alice.runningTokens.length);
+		let bobTokens = await onChainQueryA.queryTokensOwnedByAddress(
+			testenv.perunArt,
+			bob.wallet.address,
+		);
+		expect(bobTokens.value)
+			.to.have.members(bob.runningTokens)
+			.and.length(bob.runningTokens.length);
+	};
+
+	it("correctly returns the tokens in the address' possession after transfers across blocks", async function () {
+		return transferTest(true);
+	});
+
+	it("correctly returns the tokens in the address' possession after single-block transfers", async function () {
+		return transferTest(false);
+	});
+
+	afterEach(async function () {
+		// send the tokens back
+		await seededTokenTransfer(alice, bob, bTransferSeed, bob.tokens, true);
+		await seededTokenTransfer(
+			bob,
+			alice,
+			aTransferSeed,
+			alice.tokens,
+			false,
+		);
+	});
+});
--- a/src/ledger/backend/ethereumOnChainQuerier.ts
+++ b/src/ledger/backend/ethereumOnChainQuerier.ts
@@ -1,0 +1,51 @@
+import { providers, Signer, BigNumber } from "ethers";
+
+import { IERC721__factory } from ".";
+import { Tokens } from "../assets";
+
+import { OnChainQuerier } from "../onChainQuerier";
+import { TypedEvent } from "./contracts/commons";
+
+export class EthereumOnChainQuerier implements OnChainQuerier {
+	readonly provider: providers.Provider | Signer;
+
+	constructor(provider: providers.Provider | Signer) {
+		this.provider = provider;
+	}
+
+	async queryTokensOwnedByAddress(
+		token: string,
+		address: string,
+	): Promise<Tokens> {
+		const erc721 = IERC721__factory.connect(token, this.provider);
+		const outboundFilter = erc721.filters.Transfer(address, null, null);
+		const inboundFilter = erc721.filters.Transfer(null, address, null);
+		const [outbound, inbound] = await Promise.all([
+			erc721.queryFilter(outboundFilter),
+			erc721.queryFilter(inboundFilter),
+		]);
+
+		const ownershipHistory = [
+			...outbound.map((tx) => ({ ...tx, owned: false })),
+			...inbound.map((tx) => ({ ...tx, owned: true })),
+		];
+		ownershipHistory.sort((x, y): number => {
+			const blockDiff = x.blockNumber - y.blockNumber;
+			return blockDiff === 0
+				? x.transactionIndex - y.transactionIndex
+				: blockDiff;
+		});
+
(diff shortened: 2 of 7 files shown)


src/test/mocks.ts
```diff
@@ -254,3 +256,13 @@ function newTxReceiptResult(
 	const txr = new TxReceipt(tx, delta);
 	return new Result(id, txr);
 }
+
+class MockOnChainQuerier implements OnChainQuerier {
+	constructor() {}
+	async queryTokensOwnedByAddress(
+		token: string,
+		address: string,
```
ndzik, line 264:
```suggestion
		_token: string,
		_address: string,
```

src/ledger/onChainQuerier.ts
```diff
@@ -0,0 +1,6 @@
+export interface OnChainQuerier {
+	queryTokensOwnedByAddress(
+		token: string,
+		address: string,
+	): Promise<string[]>;
```
ndzik, line 5: I propose to return `Promise<Tokens>`, since it is part of our API and it makes clear what to expect.

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+
+import * as test from "#erdstall/test";
+import { Environment, setupEnv } from "test/ledger";
+import { Wallet } from "@ethersproject/wallet";
+import { IERC721Minter__factory, IERC721__factory } from ".";
+import { Tokens } from "../assets";
+import PRNG from "#erdstall/test/random";
+import { afterEach } from "mocha";
+
+const TOKEN_SIZE = 4;
```
ndzik, line 15:
```suggestion
const PRIMES = [2, 3, 5, 7];
const TOKEN_SIZE = PRIMES.length;
```

Force whoever comes after you to think about the test setup and tie implicit assumption like that together. This is in context to the way you loop over the `{a,b}TransferSeed`s below.

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+		);
+		for (const id of aliceTokens.value) {
+			await erc721Minter.mint(alice.address, id);
+		}
+		for (const id of bobTokens.value) {
+			await erc721Minter.mint(bob.address, id);
+		}
+	});
+
+	beforeEach(async function () {
```
ndzik, line 52: This is a big function and the striking similarity between the actions each user takes suggests that we should probably extract a common function.

Maybe it is just my brain not being up to the task, at least I had to factor the functions to make sure I understand how the tracking really works. I might just have been overwhelmed (:

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+		for (const bobToken of bobTokens.value) {
+			bobRunningTokens.push(bobToken.toString());
+		}
+		let ierc721 = IERC721__factory.connect(testenv.perunArt, alice);
+		for (const [index, value] of [
+			aTransferSeed % 2,
+			aTransferSeed % 3,
+			aTransferSeed % 5,
+			aTransferSeed % 7,
+		].entries()) {
```
ndzik, line 69: Maybe:
```suggestion
		for (const [index, value] of shuffleWithSeed(aTransferSeed)) {
```

With a function like:

```ts
	const shuffleWithSeed = (seed: number) => {
		return Array.from(
			{ length: TOKEN_SIZE },
			(_, i) => seed % PRIMES[i],
		).entries();
	};
```

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+					aliceTokens.value[index],
+				);
+				bobRunningTokens.push(aliceTokens.value[index].toString());
+				aliceRunningTokens.splice(
+					aliceRunningTokens.indexOf(
+						aliceTokens.value[index].toString(),
+					),
+					1,
+				);
+			}
```
ndzik, line 83: Just a general remark: We always _try_ to keep the indentation as low as possible, makes the code easier to read:

```suggestion
			if (value % 2 !== 0) {
				continue;
			}
			await ierc721.transferFrom(
				alice.address,
				bob.address,
				aliceTokens.value[index],
			);
			bobRunningTokens.push(aliceTokens.value[index].toString());
			aliceRunningTokens.splice(
				aliceRunningTokens.indexOf(aliceTokens.value[index].toString()),
				1,
			);
```

Nitpick: Personally, I would extract the "transfer&caching" logic into its own function like this:

```ts
	const transferNFTAndTrack = async (
		erc721: IERC721,
		index: number,
		[sender, sTokens, sCache]: [string, Tokens, string[]],
		[recipient, rCache]: [string, string[]],
	) => {
		await erc721.transferFrom(sender, recipient, sTokens.value[index]);
		rCache.push(sTokens.value[index].toString());
		sCache.splice(sCache.indexOf(sTokens.value[index].toString()), 1);
	};
```

which results in this, IMO easier to digest, setup code:

```ts
		let ierc721 = IERC721__factory.connect(testenv.perunArt, alice);
		for (const [index, value] of shuffleWithSeed(aTransferSeed)) {
			if (value % 2 !== 0) {
				continue;
			}
			await transferNFTAndTrack(
				ierc721,
				index,
				[alice.address, aliceTokens, aliceRunningTokens],
				[bob.address, bobRunningTokens],
			);
		}
```

ndzik: The same applies to `bob`s setup.

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+
+	afterEach(async function () {
+		// send the tokens back
+		let ierc721 = IERC721__factory.connect(testenv.perunArt, bob);
+		for (const [index, value] of [
+			aTransferSeed % 2,
+			aTransferSeed % 3,
+			aTransferSeed % 5,
+			aTransferSeed % 7,
+		].entries()) {
```
ndzik, line 130:
```suggestion
		for (const [index, value] of shuffleWithSeed(aTransferSeed)) {
```

ndzik: You get what I mean.

ndzik (changes requested):

src/ledger/backend/ethereumOnChainQuerier.spec.ts
ndzik: Instead of my proposed `shuffleWithSeed` function you can of course just map over the existing `PRIMES` array, in case you take that route.

ndzik (approved):
Possibly the cleanest I have ever felt. THIS is the fix I needed to get me through the day.
Just overall clean. Now, I feel the desire to lift a :truck: and eat a :horse:

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+	};
+
+	const seededTokenTransfer = async (
+		from: Entity,
+		to: Entity,
+		seed: number,
+		tokenPool: Tokens,
+	): Promise<void> => {
+		const ierc721 = IERC721__factory.connect(testenv.perunArt, from.wallet);
+		for (let index = 0; index < TOKEN_SIZE; ++index) {
```
sebastianst, line 60: Consider using `PRIMES.forEach((p, i) => {...})`.
I'd also define `const tknId = tokenPool.value[i]` because you use this expression three times inside the loop.

sebastianst (changes requested):

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -0,0 +1,54 @@
...
+				const nextToken = inbound[0].args.tokenId.toString();
+				tokenOwnedMemo[nextToken] = true;
+				inbound.shift();
+			}
+		}
+
+		return new Tokens(
+			Object.keys(tokenOwnedMemo)
+				.filter((key) => tokenOwnedMemo[key])
+				.map(BigInt),
```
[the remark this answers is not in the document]
Ruben-Kruepper: This is exactly why I didn't use a map. Norbert asked the same question. I find it much harder to understand what is going on here. When rewriting the algorithm I took the opportunity to work in a map in a way I find easy to understand though. :smile:

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -0,0 +1,54 @@
...
+			erc721.queryFilter(outboundFilter),
+			erc721.queryFilter(inboundFilter),
+		]);
+
+		outbound.sort((x, y) => x.blockNumber - y.blockNumber);
+		inbound.sort((x, y) => x.blockNumber - y.blockNumber);
+
+		const tokenOwnedMemo: { [key: string]: boolean } = {};
+
+		while (outbound.length > 0 || inbound.length > 0) {
```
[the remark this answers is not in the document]
Ruben-Kruepper: This is what happens when you start fixing an algorithm instead of taking a break and starting fresh. :sob: 

I agree, this is easier to understand and equally performant.

src/ledger/backend/ethereumOnChainQuerier.spec.ts
Ruben-Kruepper: I got used to not using `forEach` while developing JS backend because of the performance implications. In case you ever test hundreds of thousands of transactions: You have been warned :warning:

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+			entity.runningTokens.push(token);
+		}
+
+		const erc721Minter = IERC721Minter__factory.connect(
+			testenv.perunArt,
+			entity.wallet,
+		);
+		for (const id of entity.tokens.value) {
+			await erc721Minter.mint(entity.wallet.address, id);
+		}
```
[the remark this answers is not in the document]
Ruben-Kruepper: Sadly, this doesn't work. The test environment won't allow parallel execution like this. This is probably why this was done sequentially in another test, too, which I read to write my own.

src/ledger/backend/ethereumOnChainQuerier.spec.ts
Ruben-Kruepper: And our test env doesn't play nice with multiple async transfer calls, so this won't work either.

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -24,31 +25,43 @@ export class EthereumOnChainQuerier implements OnChainQuerier {
...
+				[string, string, BigNumber] & {
+					from: string;
+					to: string;
+					tokenId: BigNumber;
+				}
+			> {}
+
+		const compareTransactions = (
+			x: transferEvent,
+			y: transferEvent,
```
sebastianst, line 39: If you do
```suggestion
			x: { blockNumber: number; transactionIndex: number },
			y: { blockNumber: number; transactionIndex: number },
```
, like I suggested, you can delete lines 28-25 (`interface transferEvent`) and cleanup some imports. Now that you use this function only once, you could even pass it directly to `ownershipHistory.sort(...)` and wouldn't need to specify the types of `x` and `y`.

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -24,31 +25,43 @@ export class EthereumOnChainQuerier implements OnChainQuerier {
...
+		ownershipHistory.sort(compareTransactions);
+
+		const tokenOwnedMemo = new Map<bigint, boolean>();
+
+		for (let i = 0; i < ownershipHistory.length; i++) {
+			tokenOwnedMemo.set(
+				ownershipHistory[i].args.tokenId.toBigInt(),
+				ownershipHistory[i].owned,
+			);
 		}
```
sebastianst: Why not 
```suggestion
		for (const h of ownershipHistory) {
			tokenOwnedMemo.set(
				h.args.tokenId.toBigInt(),
				h.owned,
			);
		}
```
?

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -1,9 +1,10 @@
-import { providers, Signer } from "ethers";
+import { providers, Signer, BigNumber } from "ethers";
 
 import { IERC721__factory } from ".";
-import { Tokens } from "../assets";
+import { mapNFTs, Tokens } from "../assets";
```
sebastianst, line 4:
```suggestion
import { Tokens } from "../assets";
```
the import is not used.

sebastianst (changes requested):

src/ledger/backend/ethereumOnChainQuerier.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+		expect(aliceTokens.value).to.have.length(alice.runningTokens.length);
+		let bobTokens = await onChainQueryA.queryTokensOwnedByAddress(
+			testenv.perunArt,
+			bob.wallet.address,
+		);
+		expect(bobTokens.value).to.have.members(bob.runningTokens);
+		expect(bobTokens.value).to.have.length(bob.runningTokens.length);
+	});
+
+	it("correctly returns the tokens in the address' possession after single-block transfers", async function () {
```
sebastianst, line 128: If I see correctly, this test and the prior are the exact same, besides the boolean `forwardChain` being set to true and false. In such cases, please always move the code into a function and then call it with the different parameters (here, `true` and `false`). D.R.Y.

src/ledger/backend/ethereumOnChainQuerier.ts
```diff
@@ -0,0 +1,60 @@
...
+		]);
+
+		interface transferEvent
+			extends TypedEvent<
+				[string, string, BigNumber] & {
+					from: string;
+					to: string;
+					tokenId: BigNumber;
+				}
+			> {}
```
sebastianst, line 35: remove, it is unused.

sebastianst (changes requested):

sebastianst (approved):
Waiting for #117 to be merged

sebastianst (approved):

PR #114, opened 2021-11-02
Use lowercase addresses for JSON encoding values

ndzik:


src/ledger/assets/assets.spec.ts
```diff
@@ -5,13 +5,22 @@ import { describe, it } from "mocha";
...
-		const vs = Assets.fromJSON(JSON.parse(testAssets));
+		const jsonAssetsWithChecksum: { [key: string]: any } = JSON.parse(
+			testAssetsWithChecksum,
+		);
+		const jsonAssetsNoChecksum = Object.fromEntries(
+			Object.entries(jsonAssetsWithChecksum).map(([k, v]) => [
+				k.toLowerCase(),
+				v,
+			]),
+		);
```
sebastianst, line 21: nifty

src/ledger/assets/assets.ts
```diff
@@ -36,7 +37,11 @@ export class Assets implements ABIValue {
 	static fromJSON(data: any): Assets {
```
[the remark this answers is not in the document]
ndzik: Yes, I remember thinking about adding a safety call everywhere. Then I decided to pass that responsibility to the user.
Thinking about it again, it is probably a saner choice to add that extra call.

The `ctor` uses `addAsset` internally, so this is checked.
`hasAsset` now also does a check.
`addAsset` as well.

This also made me check the whole project again and I found **even** more uses of `Address`es or addresses with the type `string` as map-keys.

src/ledger/backend/tokencache.ts
```diff
@@ -188,7 +192,7 @@ export class TokenFetcher implements TokenProvider {
 		ttype: TokenType,
 		token: string | Address,
 	): Responder {
-		token = token instanceof Address ? token.toString() : token;
+		token = token instanceof Address ? token.key : token.toLowerCase();
```
sebastianst, line 195: This is the same code as the `private ensureAddrKey` in the other commit. Maybe make `ensureAddrKey` an exported function in the same package as `Address`? Maybe also rename it to `addressKey` then ^^ And maybe use it instead of "manual" `addr.toLowerCase()` calls. If we use `addressKey` function everywhere, should we ever want to change the key coding (which we hopefully don't), we don't need to search the whole codebase for `toLowerCase()` calls.

sebastianst (approved):

PR #115, opened 2021-11-04
test/ledger/env: Allow configuring underlying ganache instance

ndzik:
This version just exposes the raw ganache configuration struct to be passed as an optional parameter to the setup function of the testenvironment.

There is an argument to be made regarding the _complexity_ of this usage for someone unfamiliar with the stack. It would be easier, but more limiting, to just allow passing a flag like `"automine" | "mineOnTX"` or similar but doing so fixes the blocktime to something **WE** define in our test package. If this also would have to be configurable we would need to expose this too and all other configurations might leak out one by one over time.

This is my argument for this straight forward approach.

EDIT:
Regarding those  `as any` casts after using `import` statements for the contract ABIs: `tsc` has difficulties inferring the correct `abi: any[]` type for the `SimpleContractJSON` type (which is not exported by `ethereum-waffle` :angry: ), thus forcing us to do it like this...

### DIFF
--- a/src/test/ledger/env.ts
+++ b/src/test/ledger/env.ts
@@ -19,6 +19,10 @@ import erc721holderABI from "../../ledger/backend/contracts/abi/ERC721Holder.jso
 import ethholderABI from "../../ledger/backend/contracts/abi/ETHHolder.json";
 import perunArtABI from "../../ledger/backend/contracts/abi/PerunArt.json";
 
+// ethereum-waffle does not expose this type...
+type _params = ConstructorParameters<typeof MockProvider>;
+export type MockProviderOptions = NonNullable<_params[0]>["ganacheOptions"];
+
 export interface Environment {
 	provider: providers.Web3Provider;
 	perun: string;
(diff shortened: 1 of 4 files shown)


src/test/ledger/env.ts
```diff
@@ -38,25 +43,25 @@ const PERUNART_NAME = "PerunArt";
...
 const OP = 0,
 	TEE = 1;
 
 export async function setupEnv(
 	numOfPrefundedAccounts: number = 1,
 	epochDuration: number = 3,
-	lprovider?: providers.Web3Provider,
-	lop?: Wallet,
+	ganacheOptions?: MockProviderOptions,
 	pacc?: Wallet,
```
sebastianst, line 53: Does it even still make sense to pass `pacc?: Wallet`? Or does it cause inconsistencies?

This also now removes the ability to inject a `lprovider`? Do we use this in any other project?

ndzik: Taken from my commit message:

Also moved global variables into `setupEnv` function:
  * In hindsight unnecessary.
  * Passing a local provider can be helpful, but never was so far.
  * With the `lprovider` gone there is also no need for `lop` to be
    passed to the setup function.
    
Initially I did not remove `pacc` because it was independent from the other settings I deleted.

Thinking about it again, it **HAS** to be removed, because it effectively overrides the first `numOfPrefundedAccounts` option a user passes. So lets just purge this too.

sebastianst: Sounds good. We better remove it and if we ever need this feature, we can add it, knowing how exactly.

sebastianst (approved):
Golden 🏅

PR #117, opened 2021-11-08
Speed up testsetup

ndzik:
* Speeds up the setup of our test environment
* Ensures the setup finishes with a new epoch starting
* Fixes an oversight in `ethwrapper.spec` where I used `currentEpoch - 1n` to challenge instead of `currentEpoch - 2n`, which only worked because the setup finished in such a way, that the next TX triggered a new epoch :disappointed:

### DIFF
--- a/src/test/ledger/env.ts
+++ b/src/test/ledger/env.ts
@@ -212,6 +212,9 @@ export async function setupEnv(
 		return bdelta;
 	};
 
+	// seal current epoch!
+	sealEpoch(await currentEpoch());
+
 	return {
 		provider: provider,
 		erdstall: contracts[erdstallContract].address,
--- a/src/test/ledger/env.ts
+++ b/src/test/ledger/env.ts
@@ -58,6 +58,13 @@ export async function setupEnv(
 	const provider = new MockProvider(
 		ganacheOptions ? { ganacheOptions: ganacheOptions } : undefined,
 	);
+
+	const mineBlocks = async (n: number | bigint = 1) => {
+		for (let i = 0; i < n; i++) {
+			await provider.send("evm_mine", []);
+		}
+	};
+
 	const wallets = provider.getWallets();
 
 	const op = wallets[OP];
@@ -85,10 +92,11 @@ export async function setupEnv(
 
 	const deployAndStore = async (deployments: typeof contractDeployments) => {
 		for (const [abi, args] of deployments) {
-			const contract = await deployContract(op, abi, args, {
+			const contract = deployContract(op, abi, args, {
 				nonce: nonce++,
 			});
-			contracts.push(contract);
+			await mineBlocks(8);
+			contracts.push(await contract);
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


src/test/ledger/env.ts
```diff
@@ -85,10 +92,11 @@ export async function setupEnv(
 
 	const deployAndStore = async (deployments: typeof contractDeployments) => {
 		for (const [abi, args] of deployments) {
-			const contract = await deployContract(op, abi, args, {
+			const contract = deployContract(op, abi, args, {
 				nonce: nonce++,
 			});
-			contracts.push(contract);
+			await mineBlocks(8);
```
sebastianst, line 98: You inserted a couple of `await mineBlocks(x)` calls. Shouldn't you first determine from the passed ganache options if that's even necessary and only call `mineBlocks` then?

The final `sealEpoch` call probably always makes sense though.

sebastianst: In particular here, do you need to mine a block after each contract deployments because we risk that not all contract deployments fit into a single block?

src/test/ledger/env.ts
```diff
@@ -114,6 +122,7 @@ export async function setupEnv(
 		const tx = await call(arg1, arg2, {
 			nonce: nonce++,
 		});
+		await mineBlocks(2); // one block containing TX + 1 confirmation.
```
sebastianst, line 125: Why is it important that each token registration is in a separate block? Don't the `mineBlocks()` call after `addMinter` plus the final `sealEpoch` call suffice?

src/test/ledger/env.ts
ndzik: I just did not deem it necessary to switch the deployment behavior depending on whether or not the ganache is set to automine or autocommit. This way it works independent from the ganache configuration.

I would also think it is cleaner to only mine the blocks if a blocktime is specified.

And you are right regarding the extra mine calls for each contract. I did not want to risk that contract deployments exceed the gaslimit of a single block.

src/test/ledger/env.ts
ndzik: Just wanted to keep the error handling which was already in place. I am not sure whether or not the additional error handling is superfluous or not. Maybe you can put the final nail in the coffin and give your opinion. Putting them all in a single block is possible, I just do not think that the error output (in case of an error) is easily digestible since ganache errors are already ugly on their own :see_no_evil:

src/test/ledger/env.ts
sebastianst: Fair enough, let's keep the logic simple 👍🏻

src/test/ledger/env.ts
sebastianst: OK, if it helps debugging ganache output, then I'm fine with that ;) But wouldn't it still suffice to mine a single block? Because now you're mining one block containing two txs, and then another one containing zero txs.

ndzik: You are correct, I was under the impression that `tx.wait(numOfConfirmations = 1)` meant that one **more** block had to be mined for `tx.wait(1)` to return (and not wait `blockTime` seconds).

But this is not the case, so we can remove the `2`. Thanks.

sebastianst (approved):
Looks good! We will adapt when https://github.com/trufflesuite/ganache/issues/1540 is resolved.

PR #118, opened 2021-11-08
doc: Minimal documentation in TSDoc style

ndzik:


sebastianst (changes requested):
Please also add documentation for classes `Address` and `Account`.

package.json
```diff
@@ -32,10 +32,17 @@
 		"test:integration": "env TS_NODE_PROJECT=\"tsconfig.e2e.json\" mocha -b --timeout 60000 -r jsdom-global/register -r ts-node/register -r tsconfig-paths/register 'src/e2e/**/*.spec.ts'",
 		"test": "env TS_NODE_PROJECT=\"tsconfig.test.json\" mocha --timeout 15000 -r jsdom-global/register -r ts-node/register -r tsconfig-paths/register 'src/{,!(e2e)/**}/*.spec.ts'",
 		"prepack": "yarn build && ts-node ./scripts/publish_posthook.ts ./package.json ./dist/package.json",
-		"devpub": "yarn prepack"
+		"devpub": "yarn prepack",
+    "doc": "typedoc ./src ./src/api ./src/enclave ./src/ledger ./src/ledger/backend ./src/test ./src/utils"
```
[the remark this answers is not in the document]
ndzik: :eagle: :eyes: 
Did we not add a CI step which checks for formatting with `prettier`? Or was that only in NERD?

package.json
```diff
@@ -32,10 +32,17 @@
...
 	},
 	"repository": "github:perun-network/erdstall-ts-sdk",
-	"keywords": ["blockchain", "ethereum", "plasma", "scaling", "TypeScript"],
+	"keywords": [
+		"blockchain",
+		"ethereum",
+		"plasma",
+		"scaling",
+		"TypeScript"
+	],
```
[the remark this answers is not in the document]
ndzik: Apparently it was my prettier, but I am confused as to why it did not use tabs instead of spaces before...

package.json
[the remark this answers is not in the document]
sebastianst: We actually don't check it. Our [CI is quite minimal](https://github.com/perun-network/erdstall-ts-sdk/blob/main/.github/workflows/ci.yml) at the moment ;) Created a new issue https://github.com/perun-network/erdstall-ts-sdk/issues/120

package.json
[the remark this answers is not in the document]
ndzik: My `prettier` tells me:

```
[~/projects/git/erdstall-ts-sdk] >>> prettier -c package.json
Checking formatting...
All matched files use Prettier code style!
```

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
```
sebastianst, line 11: why not `ubuntu-latest`?

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,29 @@
...
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
```
sebastianst, line 13: what's this?

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "16"
+          cache: "yarn"
+      - name: Install dependencies
+        run: yarn install
```
sebastianst, line 22:
```suggestion
        run: yarn
```

src/ledger/account.ts
```diff
@@ -8,6 +8,10 @@ import {
 } from "#erdstall/export/typedjson";
 import { Assets } from "#erdstall/ledger/assets";
 
+/**
+ * Account implements the state an account within Erdstall might have and is
+ * returned by an `EnclaveReader.getAccount` query.
```
sebastianst, line 13:
```suggestion
 * Account is the state of a user within Erdstall, including the last nonce and free and locked assets. It is
 * returned by an `EnclaveReader.getAccount` query.
```

.github/workflows/gh-pages.yml
sebastianst: Is it somehow possible to only install the subset of dependencies that are required for generating the docs?

.github/workflows/gh-pages.yml
ndzik: [concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency)

As far as I understand it just avoids the same workflow to be run at the same time, e.g. workflow triggered by a merge into `main` and someone manually requesting the CD workflow.

We can remove it though, if you say this is overkill/unnecessary here.

.github/workflows/gh-pages.yml
ndzik: I definitely did not copy paste this from their example :clown_face:

.github/workflows/gh-pages.yml
ndzik: Not that I am aware of, though something like `npx` **could** help. Would only fetch the required program and execute it.

ndzik: Interesting, the dependencies are required because `TypedDoc` seems to be utilizing `tsc` under the hood, which itself requires all bindings to exist.

.github/workflows/gh-pages.yml
ndzik: I removed it now, you can decide whether you seem it fits or not. Since we do not trigger the workflow with pull requests it might really be superfluous.

sebastianst: No it's a good idea to use it if it avoids duplicate runs. We have a limited number of free CI minutes per month ;)

.github/workflows/gh-pages.yml
sebastianst: OK makes sense. You could still remove the ` install`

.github/workflows/gh-pages.yml
sebastianst: Inspired me to create https://github.com/perun-network/erdstall-ts-sdk/issues/123

sebastianst: However, for `gh-pages` it is probably superfluous because, as you said, it would only be triggered manually or by a merge into main, both very conscious actions that shouldn't happen accidentally at the same time.

sebastianst (approved):

PR #119, opened 2021-11-09
Remove `Stages`

Ruben-Kruepper:


src/ledger/backend/connection.spec.ts
```diff
@@ -47,13 +48,28 @@ describe("ErdstallConnection", () => {
...
-			const ctx = await stage.value;
-			const rec = await ctx.wait();
+		const [stages, numStages] = await conn.deposit(assets);
+		for await (const [name, stage] of stages) {
+			const rec = await stage.wait();
 			expect(rec.status, "depositing should have worked").to.equal(0x1);
 		}
 
+		let expectedNumStages: number = 0;
+		assets.values.forEach((value, key) => {
```
ndzik, line 58: I would prefer a more generic implementation which does not require previous knowledge about the type of contracts used, similar to this:

```ts
function depositStagesForAssets(a: Assets): number {
	return Array.from(a.values.entries())
		.map(([token, asset]) => {
			switch (asset.typeTag()) {
				case TypeTags.Amount:
					// ETHZERO -> one onchain transfer.
					// ERC20 -> one onchain approve + one onchain transfer.
					return token === ETHZERO ? 1 : 2;
				case TypeTags.Tokens:
					// ERC721 -> Tokens.length onchain approves + one onchain transfer.
					return (asset as Tokens).value.length + 1;
			}
		})
		.reduce((a, b) => a + b);
}
```

Adjust to your liking, but this way you do not have to bother with `addressKey`s at all, since `ETHZERO` is `0x00000...` which is the same no matter if checksum encoded or not.

src/erdstall.ts
```diff
@@ -57,21 +57,36 @@ export interface Trader {
 export interface Depositor {
 	deposit(
 		assets: Assets,
-	): Promise<Stages<Promise<ethers.ContractTransaction>>>;
+	): Promise<
+		[
+			AsyncGenerator<[StageName, ethers.ContractTransaction], void, void>,
+			number,
+		]
+	>;
```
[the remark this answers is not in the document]
ndzik: Optimally the caller then would just use the return value using the destructuring assignment for objects, s.t. the following codechange only has to apply (given that the member names match):

```ts
		const [stages, numStages] = await conn.deposit(assets);
```
to
```ts
		const {stages, numStages} = await conn.deposit(assets);
```
The stage itself (`[StageName, ContractTransaction]`) can stay a tuple, since the type is expressive enough on its own IMHO.

ndzik (changes requested):
Deleted code is the best code :tada: :rocket: :sparkles:

ndzik (approved):
I approve.

src/erdstall.ts
```diff
@@ -153,6 +153,11 @@ export interface Trader {
 	acceptTrade(offer: TradeOffer): Promise<TxReceipt>;
 }
 
+export interface StagesGenerator {
+	stages: AsyncGenerator<[StageName, ethers.ContractTransaction], void, void>;
+	numStages: number;
+}
+
```
ndzik, line 160: Maybe move to `utils/staged_promise.ts` instead?

ndzik (commented):
Just a comment.

src/ledger/backend/writeconn.ts
```diff
@@ -68,56 +68,26 @@ export class LedgerWriteConn extends LedgerReadConn implements LedgerWriter {
 			calls.push(...depositCalls);
 		}
 
-		return this.call(calls);
+		return { stages: this.call(calls), numStages: calls.length };
 	}
 
-	private async call(
+	private async *call(
 		calls: DepositCalls,
```
sebastianst, line 75: I think the type `DepositCalls` needs to be renamed. It's not only `DepositCalls` but calls in general, right?

src/erdstall.ts
```diff
@@ -153,6 +153,11 @@ export interface Trader {
 	acceptTrade(offer: TradeOffer): Promise<TxReceipt>;
 }
 
+export interface StagesGenerator {
+	stages: AsyncGenerator<[StageName, ethers.ContractTransaction], void, void>;
```
sebastianst, line 157: What's actually the purpose of the `StageName` parameter? Can't we just do
```suggestion
	stages: AsyncGenerator<ethers.ContractTransaction, void, void>;
```
? The caller either calls `client.deposit` or `client.withdraw` etc and knows, what he's calling.

src/utils/staged_promise.ts
```diff
@@ -2,73 +2,3 @@
 "use strict";
 
 export type StageName = "approve" | "deposit" | "withdraw";
```
sebastianst, line 4: See other comment, I think we can remove everything related to `StageName`.

src/ledger/backend/connection.spec.ts
```diff
@@ -28,6 +34,24 @@ describe("ErdstallConnection", () => {
 	let contract: Erdstall;
 	let conn: LedgerWriter;
 
+	const depositStagesForAssets = (assets: Assets): number => {
```
sebastianst, line 37: In the spirit or renaming the generic `Stages` to the more specific term `Transactions`, this should be renamed to something like
```suggestion
	const numDepositTxs = (assets: Assets): number => {
```

sebastianst (changes requested):

src/ledger/backend/writeconn.ts
sebastianst: Same for type `DepositCall` in file `tokenmanager.ts`. Looks like a generic `Call` to me.

src/erdstall.ts
sebastianst: @ndzik explained to me that it's about differentiating the different calls during deposits. Wouldn't it then actually make sense to replace `StageName` by just a `string` and then write something more specific into it, like `erc721/${tokenAddr}/{id}/approve` or `erc20/${tokenAddr}/approve` etc? Because if deposit is called on multiple assets, only knowing that it's an `approval` or `deposit` call is ambiguous. Anyway, we can think about this in a later issue.

sebastianst (approved):
Just, wow.

PR #125, opened 2021-11-24
test/env: Decrease pollingInterval on Provider

ndzik:
Speeds up tests from ~50s to ~17s.

I also tried using `evm_mine` calls like it is done in the testenv setup but this turned out to be highly inconsistent because any change to the tests (ordering, new testcases etc.) seemed to influence to minimum number of blocks needed for this **hack** to work. So I went with simply reducing the `provider.pollingInterval`.


sebastianst (approved):
Extraordinary 🌟

PR #127, opened 2021-11-25
CI: Add integration test & prettier step

ndzik:


.github/workflows/ci.yml
```diff
@@ -11,15 +11,38 @@ jobs:
...
+          path: "erdstall"
+      - name: "Set up Go 1.x"
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.17
+      - name: "Build walkthrough operator and store in environment"
+        working-directory: "erdstall"
+        run: |
+          go build -o wo ./cmd/walkthrough/main.go
+          echo "ERDSTALL_EXE=$( realpath wo )" >> $GITHUB_ENV
```
sebastianst, line 35: I think you should move the whole Go stuff just before running `Test:Integration`. If any of the next steps fail, the whole Go setup was a waste.

And maybe we should even put the End-2-End Test into a separate `job`?

.github/workflows/ci.yml
```diff
@@ -11,15 +11,38 @@ jobs:
...
+      - name: "Install dependencies"
         run: yarn install
-      - name: Build
+      - name: "Build"
         run: yarn build
-      - name: Test
+      - name: "Test"
         run: yarn test
+      - name: "Test:Integration"
+        run: ERDSTALL_EXE=${{ env.ERDSTALL_EXE }} yarn test:integration
```
sebastianst, line 48: I suggest we rename them, `Integration` is not very precise. Like
```suggestion
      - name: "End-2-End Test" // or "Walkthrough"
        run: ERDSTALL_EXE=${{ env.ERDSTALL_EXE }} yarn test:e2e // or test:walkthrough
```

.github/workflows/ci.yml
sebastianst, line 35: Why so complicated? Can't you just do `ERDSTALL_EXE=erdstall/wo yarn test:integration` below?

sebastianst (changes requested):

.github/workflows/ci.yml
ndzik: That is a very good idea!

ndzik: That will work too. I just prefer absolute paths, because of relative paths always being a security nightmare in bash scripts.

But I guess it should not be a problem here.

ndzik: As to whether or not we want a separate `job` for the end to end test:
Would this not require us to use something like `actions/upload-artifact` to avoid repulling all dependencies and rebuilding the project again or is there an easier way to share the data between the jobs?

If it is only doable by the mentioned gh action script then I would argue it is not really worth it to split. We could also argue about the possibility that the end2end test should stand completely for itself, s.t. it can be triggered on its own?

sebastianst: We discussed to keep it sequentially in a single job.

sebastianst (approved):

PR #129, opened 2021-11-30
Log seed on fail

Ruben-Kruepper:


src/api/responses/balanceproof.spec.ts
```diff
@@ -43,4 +43,10 @@ describe("BalanceProofs", function () {
 		return expect(erd.verifyBalance(bal.asABI(), bp.sig.value)).to.not.be
 			.reverted;
 	});
+
+	afterEach(function () {
+		if (this.currentTest && this.currentTest.state === "failed") {
+			console.log(`Failed with PRNG ESSEED ${rng.seed}`);
+		}
```
sebastianst, line 50: Please put this into a function `logSeedIfFailed(rng)` that you export from `test`, then you only need to call `afterEach(() => { logSeedIfFailed(rng); });`. Always avoid repetition where possible.

sebastianst (changes requested):

src/api/responses/balanceproof.spec.ts
Ruben-Kruepper: Sadly this isn't possible; this would have been my preferred solution, too. `this` isn't scoped properly when defining the callback as a function somewhere else. A regular function definition creates its own `this`.  `mocha` must be working some magic under the hood.

sebastianst: Ah I see, you gotta do
```suggestion
	afterEach(function () { logSeedIfFailed(rng); });
```
of course for that reason.

You could also use a generator `function seedLogger(rng) { return function() { ... } }` and then call `afterEach(seedLogger(rng))`.

sebastianst (approved):

PR #130, opened 2021-12-06
transactions/trade: Add SignedTradeOffer Closes #98

### LINKED ISSUE #98: Create type `SignedTradeOffer` and let `TradeOffer.sign` return it
Field `sig` of `SignedTradeOffer` should be a required property (is optional in `TradeOffer`).

If possible, creation of a copy should be avoided.

This can be used to improve the type safety of trade offer handling in NERD's `MarketplaceServer`.

Ruben-Kruepper:


sebastianst:
Like mentioned in the original issue, the idea was to ultimately improve the type safety in the backend server. For the trading API, trade offers are saved into a mapping of `TradeOffer`s. Then when a trade offer is executed, it it currently asserted that the offer has a sig. Please check if this can be improved, so that incoming trade offers are checked for non-nil sigs and then saved into a mapping of `SignedTradeOffer`s.

PR #134, opened 2022-02-16
Adaptation to new transaction protocol

Ruben-Kruepper:
Adapt the SDK to the new transaction protocol. This will apparently require more structural changes to facilitate sufficient performance.


src/session.ts
```diff
@@ -46,6 +49,11 @@ export class Session extends Client implements ErdstallSession {
 		this.enclaveWriter.on("error", () => {
 			this.nonce = 0n;
 		});
+		this.receiptDispatcher = new ReceiptDispatcher(this.erdstallConn);
+		this.on(
+			"receipt",
+			this.receiptDispatcher.watch.bind(this.receiptDispatcher),
+		);
```
ndzik, line 56: We still have to brainstorm and think if we come up with a better architecture that does not require us to harden our logic against a users desire to clean stuff up...

src/test/mocks.ts
```diff
@@ -264,7 +264,7 @@ function newTxReceiptResult(
 ): Result {
 	const _acc = acc ? acc : new Account(tx.nonce, new Assets(), new Assets());
 	const delta = new Map<string, Account>([[tx.sender.key, _acc]]);
-	const txr = new TxReceipt(tx, delta);
+	const txr = new TxReceipt(tx, delta, 1);
```
ndzik, line 267: Give the function `newTxReceiptResult` a default argument for `TXStatus` and use it here.

ndzik (changes requested):

src/api/responses/txreceipt.ts
```diff
@@ -19,11 +19,22 @@ export class TxReceipt extends ErdstallObject {
 
 	@jsonMapMember(String, () => Account, { shape: MapShape.OBJECT })
 	delta: Map<string, Account>;
+	@jsonMember(Number) status: number;
```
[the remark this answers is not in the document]
Ruben-Kruepper: Status is a number and the mapping was not yet fixed at the time of writing. Has this changed?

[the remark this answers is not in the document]
ndzik: They are fixed now.

src/session.ts
```diff
@@ -166,9 +183,23 @@ export class Session extends Client implements ErdstallSession {
...
 		const tx = new Trade(this.address, await this.nextNonce(), offer);
 		await tx.sign(this.erdstallConn!.erdstall(), this.signer);
-		return this.enclaveWriter.trade(tx);
+		const hash = tx.transactionHash(this.erdstallConn.erdstall());
+		const pendingReceipt = this.receiptDispatcher.register(hash);
+		const pendingAccepted = this.enclaveWriter.trade(tx);
+		return { pendingReceipt, pendingAccepted };
+	}
+
+	removeAllListeners(): void {
```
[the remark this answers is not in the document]
Ruben-Kruepper: This life chose me.

src/api/responses/txreceipt.ts
```diff
@@ -13,17 +13,31 @@ import {
 
 const txReceiptTypeName = "TxReceipt";
 
+export enum TXStatusCodes {
```
ndzik, line 16: I'd prefer `TxStatusCode` since it will then be used as `TxStatusCode.Fail` and this reads better.

Also `Tx` because our transaction types also always start like this.

src/api/responses/txreceipt.ts
```diff
@@ -13,17 +13,31 @@ import {
...
+	Success = 1,
+}
+
 @jsonObject
 export class TxReceipt extends ErdstallObject {
 	@jsonMember(() => Transaction) tx: Transaction;
 
 	@jsonMapMember(String, () => Account, { shape: MapShape.OBJECT })
 	delta: Map<string, Account>;
+	@jsonMember(Number) status: number;
```
ndzik, line 27:
```suggestion
	@jsonMember(Number) status: TxStatusCode;
```

src/api/responses/txreceipt.ts
```diff
@@ -13,17 +13,31 @@ import {
...
+	@jsonMember(String) error?: string;
 
-	constructor(tx: Transaction, delta: Map<string, Account>) {
+	constructor(
+		tx: Transaction,
+		delta: Map<string, Account>,
+		status: number,
```
ndzik, line 33:
```suggestion
		status: TxStatusCode,
```

src/e2e/test_harness.ts
```diff
@@ -187,7 +187,8 @@ export function endToEndTestHarness(sdkActions: SDKActions) {
 					token: PERUN_ADDR,
 					asset: new assets.Amount(utils.parseEther("1").toBigInt()),
 				});
-				await sessions[ALICE].transferTo(amount, sessions[BOB].address);
+				sessions[ALICE].transferTo(amount, sessions[BOB].address)
```
ndzik, line 190:
```
Sent 500 TXs in 0.033s (0.066ms/tx | 15151.515151515152Hz)
```
I wish :laughing: 

Still have to await here first. On the other hand could you confirm that the test times out when using an array of `Promise`s?? It did so on my end and this confuses me a bit... Especially with the updated logic we should not run into issues there.

src/test/mocks.ts
```diff
@@ -261,10 +261,11 @@ function newTxReceiptResult(
 	id: string,
 	tx: Transaction,
 	acc?: Account,
+	status: number = 1,
```
ndzik, line 264:
```suggestion
	status: TxStatusCode = TxStatusCode.Success,
```

ndzik (changes requested):
In `sdk_actions.ts` for the `offchainTransfer` step: Adapt the description, s.t. it fits with the current implementation. Make sure to to mention the `TXAccepted` message as well and make sure a user knows it is useful if he is only interested in a confirmation that a TX arrived.

src/e2e/test_harness.ts
ndzik: I had another look and it seems like we miss the registration of our hash again. The way we now split the control flow with `.then` chaining does not seem to work reliably. A helperfunction which constructs our `PendingTransaction`s would also be nice to avoid all the code dup when sending TXs. To guarantee sequential execution of `receiptDispatcher.register -> transfer` you probably need a third `Promise` which gets resolved as soon as `register` is done.

```ts
.then((tx) => {
  const hash = tx.hash(...);
  this.receiptDispatcher.register(hash).then(resolveReceiptPromise);
  return this.enclaveWriter.transfer(tx)
}
```

At least this is what I would try, no guarantee. Maybe you come up with a better idea.

Ruben-Kruepper: Just drop `await` to speed up your `js` bro. Trust me bro.

ndzik: Only if you are a certified engineer :clown_face:

src/utils/receipt_dispatcher.ts
```diff
@@ -3,6 +3,7 @@ import { LedgerReader, LedgerWriter } from "#erdstall/ledger";
 
 export class ReceiptDispatcher {
 	private pendingReceiptDispatchers: Map<string, (value: TxReceipt) => void>;
+	private count = 0;
```
ndzik, line 6: Never used? Remainder from a debugging session? ^^

ndzik (changes requested):
The test for `TxReceipt`s JSON en-/decoding still has to be updated. Otherwise we do not even know if they work properly.

src/session.ts
```diff
@@ -80,38 +88,43 @@ export class Session extends Client implements ErdstallSession {
...
-		);
-		await minttx.sign(this.erdstallConn.erdstall(), this.signer);
-		return this.enclaveWriter.mint(minttx);
+		const nonce = await this.nextNonce();
+		const tx = new Mint(this.address, nonce, token, id);
+		await tx.sign(this.erdstallConn!.erdstall(), this.signer);
+		const hash = tx.hash(this.erdstallConn!.erdstall());
+		const receipt = this.receiptDispatcher.register(hash);
+		const accepted = this.enclaveWriter.mint(tx);
+		return { receipt, accepted };
```
ndzik, line 114: Last comment, rest is fine: Could we create a small helper method so we do not duplicate this code in every TX function?

ndzik (approved):
w0w

src/session.ts
ndzik: Keep it like it is because it is simple and the alternative contains black magic JS regarding the scoping of `this` and might even be more cumbersome to use.

PR #135, opened 2022-03-01
e2e: Fix time measurement in transaction benchmark

RmbRT:
Now properly distinguishes between signing latency via ganache and the
actual time it takes Erdstall to process the transaction.


ndzik (approved):
Looks sound and promising. Gonna merge. Thanks for your contribution.

PR #137, opened 2022-03-16
Rework subscriptions

### LINKED ISSUE #136: Improve `Session` and `EnclaveConnection` architecture for _persistent_/_internal_ subscriptions
We still have to brainstorm and think if we come up with a better architecture that does not require us to harden our logic against a users desire to clean stuff up...

_Originally posted by @ndzik in https://github.com/perun-network/erdstall-ts-sdk/pull/134#discussion_r812131823_

**TBD**: Will edit after sketching out a possible solution

RmbRT:
Closes #136.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 "use strict";
 
-import { ErdstallEvent, ErdstallEventHandler, EnclaveEvent } from "event";
-import { ErdstallClient } from "erdstall";
+import { InternalEnclaveWatcher } from "./internalenclavewatcher";
+import { ErdstallEvent, ErdstallEventHandler, EnclaveEvent } from "./event";
+import { ErdstallClient } from "./erdstall";
 import { Erdstall__factory } from "#erdstall/ledger/backend/contracts";
 import { ClientConfig } from "#erdstall/api/responses";
 import { Enclave, isEnclaveEvent, EnclaveReader } from "#erdstall/enclave";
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -61,6 +61,20 @@ export class Client implements ErdstallClient {
 		return this.erdstallConn.getNftMetadata(token, id, useCache);
 	}
 
+	protected on_internal<T extends EnclaveEvent>(ev: T, cb: ErdstallEventHandler<T>): void {
```
ndzik, line 64: You still need to create something like an `internal_watcher` interface with the appropriate methods. THEN you can define the following: `Client.enclaveConn: EnclaveReader & internal_watcher<EnclaveEvent>`. Or however you decided to name your `internal_watcher` interface.

The most important part in my eyes is that the `internal_watcher` methods do not bleed into the public namespace of your `Session` or `Client`.

src/enclave/connection.ts
```diff
@@ -312,7 +350,22 @@ export class Enclave implements EnclaveWriter {
 	}
 
 	private onOpen(_: Event) {
+		const calls = [];
```
ndzik, line 353: You can also add a testcase in `src/enclave/connection.spec.ts`. There you have access to the `EnclaveMockProvider` which allows you to call `close` and `open` which in turn _should_ give you the ability to test  this path with ease.

ndzik (changes requested):

ndzik (approved):
Fabulous :sparkles:

PR #153, opened 2023-04-12
Update TX encoding.

RmbRT:


package.json
```diff
@@ -93,6 +93,7 @@
 		"ethereum-waffle": "^3.4.0",
 		"ethers": "^5.4.1",
 		"isomorphic-ws": "^4.0.1",
+		"json-canonicalize": "^1.0.4",
```
DragonDev1906, line 96: Any specific reason for using [`json-canonicalize`](https://www.npmjs.com/package/json-canonicalize) instead of [`canonicalize`](https://www.npmjs.com/package/canonicalize)?

From what I've seen `canonicalize` was updated more recently, doesn't have any open PRs (granted, those are from dependabot  on build dependencies), is smaller in file size and has more downloads/other projects depending on it.

To me `canonicalize` looks better maintained and still active.

DragonDev1906 (approved):

PR #154, opened 2023-08-08
Multi backend

ndzik:


.github/workflows/ci.yml
```diff
@@ -13,11 +13,11 @@ jobs:
...
-          dry: True
-          prettier_options: --config .prettierrc --check **/*.ts
+#      - name: "Prettier Check"
+#        uses: creyD/prettier_action@v4.0
+#        with:
+#          dry: True
+#          prettier_options: --config .prettierrc --check **/*.ts
       - name: "Setup Node.js"
         uses: actions/setup-node@v3
         with:
```
DragonDev1906, line 23: Replacing the line below should fix the CI, according to the error log "@nomicfoundation/edr@0.3.2", a dependency of hardhat does not like nodejs version 16:
```diff
- node-version: "16"
+ node-version: "20"
```

Not sure if that breaks something else, but you've probably tested locally with version 18 or 20 already.

DragonDev1906 (approved):

PR #156, opened 2024-05-14
Improved wire error handling & reporting.

RmbRT:
- Now uses better stack traces.
- Handles more edge cases gracefully.


DragonDev1906 (approved):
Might make sense to convert all returns in this function to be empty, thus clearly indicating that this function doesn't return something from the called functions. I find it a bit confusing when there are `return;` and `return some_function_not_returning_anything()` in the same function, even when they are functionally identical in JavaScript (as long as the function called really does not return anything).

PR #161, opened 2025-03-31
Rework: Feature parity with WP 1.

RmbRT:
- Reworked some core abstractions.
- Abstracted how multiple chains work (sessions/clients).
- Introduced more fine-grained separation of capabilities for Apps.
- Separated signer and chain provider.
- Reworked L1 transactions:
  - Abstracted L1 transactions, introduced TxSigner, TxSender
    interfaces.
  - Added WildcardTx to represent descriptions of L1 transactions.
  - Distinguished between unsigned and signed transactions and receipts.
  - Added L1 transaction batches (for multi-transaction operations).
- Made use of #private fields to reduce the attack surface.
- Added proper Substrate transaction & session support.
- Removed lots of unnecessary type constraint juggling, opting for
  concrete types and class inheritance instead.
- Reintroduced asset operations (add, sub, cmp).
- Fixed webpack problems.


DragonDev1906 (approved):

README.md
```diff
@@ -331,7 +331,7 @@ The other part of asset handling is the `ChainAssets` class. It is fairly straig
...
+Fungible tokens and NFTs can be deposited into our L2 network by calling `Session.deposit(assets)`. This call returns an `UnsignedTxBatch` that can then be signed with `Session.signTxBatch()` and then sent with `Session.sendTxBatch()`. Only after all those transactions are mined are the deposits successful. The assets will be available in Wildcard once the Wildcard TEE witnesses the finality of the blocks containing the respective transactions. On substrate finality is reached within well under a minute, but on Ethereum (PoS), finality proofs for blocks are issued in much sparse intervals. Around epoch boundaries, Ethereum finality does currently delay substrate block processing, though, as wel can only progress into a new epoch if all blocks for that epoch on all chains have been processed.
```
DragonDev1906, line 334:
```suggestion
Fungible tokens and NFTs can be deposited into our L2 network by calling `Session.deposit(assets)`. This call returns an `UnsignedTxBatch` that can then be signed with `Session.signTxBatch()` and then sent with `Session.sendTxBatch()`. Only after all those transactions are mined are the deposits successful. The assets will be available in Wildcard once the Wildcard TEE witnesses the finality of the blocks containing the respective transactions. On substrate finality is reached within well under a minute, but on Ethereum (PoS), finality proofs for blocks are issued in much sparse intervals. Around epoch boundaries, Ethereum finality does currently delay substrate block processing, though, as we can only progress into a new epoch if all blocks for that epoch on all chains have been processed.
```

DragonDev1906 (approved):

PR #162, opened 2025-07-01
WP 2 & 3 feature parity

RmbRT:
Privacy rewrite: Multi-address accounts, encryption, new message codec, new transaction types. Some cleanup.


package.json
```diff
@@ -1,8 +1,9 @@
 {
 	"name": "@polycrypt/erdstall",
-	"version": "0.3.1",
+	"version": "0.4.2",
```
DragonDev1906: What happened to 0.4.0 and 0.4.1?

src/api/calls/getaccount.ts
```diff
@@ -1,27 +1,117 @@
...
 
-	constructor(who: Address<Crypto>) {
-		super();
-		this.who = who;
-	}
+export class GetAccount extends Transaction {
+		constructor(
+			core: TxCore,
+	// fetches the encrypted secret key via DH exchange. If so, the response of the transaction is in plaintext.
+	public aes_secret: DHPair | undefined,
```
DragonDev1906, line 32: Shouldn't this be called ecdh_pk or similar?
The comment is rather confusing and type+name doen't match.
Is this for a DH key exchange for encrypting the response? Shouldn't the enclave use the shared secret for that?
I don't understand what this is for, at least not without further context from the Rest of the codebase => Improve the comment?

src/api/calls/getaccount.ts
```diff
@@ -1,27 +1,117 @@
...
+	// whether to fetch the balances, and whether to send them in plaintext. If in plaintext, or the secret key is requested, the response of the transaction is in plaintext.
+	public balances: typeof balance_fetch_modes[number] | undefined
```
DragonDev1906: Do we need to allow the user to specify that? I think it should be enough to decide that on the enclave/runner side based on whether or not a shared secret is set for the account.

src/api/calls/getaccount.ts
```diff
@@ -1,27 +1,117 @@
...
+_transactionDecoders.set(TransactionType.GetAccount, GetAccount.decode_impl);
 
-registerErdstallType(getAccountTypeName, GetAccount);
+export class GetAccount_Output {
+		constructor(
+	public nonce: bigint,
+	public id: WildcardAddress | undefined,
+	public eth: EthereumAddress | undefined,
+	public subst: SubstrateAddress | undefined,
+	public aes_secret: { tempDH: DHPK, enc: SignedMessage<AESGCMKey> } | undefined,
```
DragonDev1906, line 98: I'm not sure if it is a good idea to send this. Cryptographically it should be okay, as long as the ecdh_pubkey is signed with the users private key. But I think we should only implement a way to retrieve the secret if we **really** need it, as it increases the attack surface.

src/api/calls/getaccount.ts
```diff
@@ -1,27 +1,117 @@
...
+	// using SignedMessage for encryption here.
+	public balances: SignedMessage<ChainAssets> | undefined,
+		) {}
```
DragonDev1906, line 101: Comment should explain why in order to be useful.

I don't think we should use a specific type like SignedMessage for something completely different (encryption), which doesn't even have a signature. That will only lead to confusion in the future.

src/api/responses/balanceproof.ts
```diff
@@ -1,161 +1,124 @@
...
-}
-
-// BalanceProofs are generated by the Enclave at the end of each transaction
-// phase for each account in the Erdstall system.
 @jsonObject
-export class BalanceProofs extends ErdstallObject {
-	accounts: Map<string, BalanceProof>;
+export class SignedBalanceProof extends ErdstallObject {
+		constructor(
+	readonly sig: SignedMessage<BalanceProof>
```
DragonDev1906, line 22: `my_balance_proof.sig.message`
`my_balance_proof.sig.signature`

Wouldn't it make more sense to "embed" (to use terminology from Go) this and not have the `.sig`?
Should be possible with JS, too.

src/api/transactions/transaction.ts
```diff
@@ -2,127 +2,202 @@
...
+		{ return new StrictNonceCheck(r.u64()); }
+};
+export class PermissiveNonceCheck extends NonceCheck {
+	constructor(public nonce: bigint) { super(); }
+	override encode_impl(w: CodecWriter): void { w.u64(this.nonce); }
+	override nonceCheckType(): NonceCheckType
+		{ return NonceCheckType.Permissive }
+	static decode_impl(r: CodecReader): PermissiveNonceCheck
+		{ return new PermissiveNonceCheck(r.u64()); }
+};
```
DragonDev1906, line 72: Do we really need one class per enum variant?

src/api/transactions/transaction.ts
```diff
@@ -2,127 +2,202 @@
...
+		let sender = Address.decode(r);
+		let sig = SignedMessage.decode(r);
+		return new SignedTransaction(sender, sig);
 	}
+}
+
+registerErdstallType(transactionTypeName, SignedTransaction);
+customJSON(SignedTransaction);
+
+export class TxCore {
```
DragonDev1906, line 131: I guess this separation is for usability?
It feels a bit weird, since TxCore doesn't have any logic and Transaction doesn't have any data, so I don't really see the benefit of splitting these two.
If it is just to pass in the data to the constructor it probably makes more sense to use a typescript interface?

src/crypto/signature.ts
```diff
@@ -45,3 +68,35 @@ export abstract class Signature<B extends Crypto = Crypto> {
...
+
+export interface SigVerifier {
+	// returns undefined if the signature is invalid. Otherwise, returns the message that was signed.
+	verifySig(s: SignedMessage): Promise<Uint8Array | undefined>;
+}
+
+// Either signed or encrypted with integrity protection, depending on the implementation.
+export class SignedMessage<T = any> {
+		constructor(
+	public message: Uint8Array, // the potentially encrypted message
```
DragonDev1906, line 80: As mentioned above: I don't think it's a good idea to mix these.
A lot of messages in Erdstall are Signed but encrypted, or encrypted but not signed. I don't think we really have messages that are both. So this should probably be split into `Signed<T>` and `Encrypted<T>`, like we have it on the Rust side. That way it is clear from the typesystem whether something is signed or encrypted.
And if both are possible/allowed we should probably just use the union of typescript: `Signed<Foo> | Encrypted<Foo>`, that way it's clear that either one is expected.
And if we really need both (I doubt that's the case) we can still use `Signed<Encrypted<T>>`, which also clearly shows in which order it happens.

src/api/calls/getaccount.ts
DragonDev1906: Conceptually a Transaction is the thing you send to the operator. Which absolutely SHOULD NOT contain the diffie hellman secret key. If we need to store that together with the transaction we should probably have a separate class containing both. But please keep the data needed for transmission separate from secret key information whenever possible.

src/crypto/wildcard/dh.ts
```diff
@@ -0,0 +1,261 @@
...
+		return new AESGCMKey(await importKey(
+			"raw",
+			key,
+			{name: "AES-GCM"},
+			true,
+			["encrypt", "decrypt"]
+		));
+	}
+
+	async verifySig(sig: SignedMessage): Promise<Uint8Array | undefined> {
```
DragonDev1906, line 217: `verifySig` doesn't make much sense in the context of AES-GCM. Should probably be called `checkIntegrity` or `checkAuthenticity` if we need that separate from decrypt.

README.md
```diff
@@ -90,16 +98,14 @@ You can install the SDK from [npm](https://www.npmjs.com/package/@polycrypt/erds
...
 
-**Sub-package `test`**&emsp; Contains internal helpers for testing.
-
-**Sub-package `utils`**&emsp; Contains mostly internal helpers for various tasks, such as encoding hexadecimal numbers, call/response message matching, and more. It is not intended for direct use by third-party projects, but exposed nonetheless. One exception is the `PendingTransaction` interface: it contains two promises: `.accepted` to query that the Wildcard operator has received a L2 transaction, and `.receipt`, which is a promise to a `TxReceipt`, which lets us verify the execution of the L2 transaction and whether it succeeded or failed.
+**Sub-package `utils`**&emsp; Contains mostly internal helpers for various tasks, such as encoding hexadecimal numbers, call/response message matching, and more. It is not intended for direct use by third-party projects, but exposed nonetheless.
```
DragonDev1906, line 108: > but exposed nonetheless

Any reason for that?

package.json
RmbRT: We already released 4.0, 4.1, and 4.2. But releases don't happen via github.

src/api/responses/balanceproof.ts
RmbRT: no it is not possible to embed in JS, you can only inherit from a single class. I could directly put members in there and then construct a SignedMessage manually from it, but that would just be more code to write.

src/api/transactions/transaction.ts
RmbRT: Not strictly, but for now I did it like that. I can change this later on.

src/api/transactions/transaction.ts
RmbRT: That would also be possible, yes.

src/api/calls/getaccount.ts
RmbRT: this only sends the key if it was requested, and only sends it in encrypted form, encrypted with the temporary ECDH pair's handshake, and that request has to be signed. Seems secure enough to me.

src/api/calls/getaccount.ts
RmbRT: Well that was the only way I found so far to not cause even more refactoring work and introduces more abstractions and concepts into the codebase. But I can consider rewriting it later on.

package.json
DragonDev1906: So to summarize:
- We have published 0.4.1 and 0.4.2 to npm but forgot to update the package.json on github (did not commit to github) => It's hard/difficult to even know which commit corresponds to these versions
- 0.4.0 is completely missing (and/or was yanked from npm), because it isn't listed in the version history there.
- All changes in this PR are backwards compatible with version 0.4.2 (according to whatever backwards compatible definition we use, that's assuming we're using Minor for breaking changes while major is at 0, as done in Rust)?

As far as I know it's best practice to update the version number in the Repo in the commit on which the new release is published. In that case this would've been at 0.4.1 and stay at 0.4.1 until we decide to do the release (which should probably be a separate PR), then update this to 0.4.2 and publish to npm and create the PR at the same time.

src/api/calls/getaccount.ts
DragonDev1906: Secure: Yes
But it should still be avoided whenever possible (to the point where it might make sense to have these as separate transactions). Main reason being that (in the long run) a GetAccount request doesn't even have to touch the enclave, assuming that the data for the user is already persisted with the users encryption key, as is the plan. => Every request that wants the aes_secret has an additional resource overhead in the TEE.

We can keep it in one request type if you want, though then we'll have to (at some point) decide in the runner whether this needs to be sent to the TEE.

src/api/responses/balanceproof.ts
DragonDev1906: And we don't want to inherit because?

I'm fine with both, just curious.

README.md
RmbRT: just for completeness. It makes no real difference AFAIK.

src/api/calls/getaccount.ts
RmbRT: No, this does not send the DH secret key. The thing that actually gets sent is the SignedTransaction, which only contains the public key. I could also rename the field to "fetch_encryption_key".

src/api/calls/getaccount.ts
RmbRT: Sometimes you want to fetch the balances in plaintext but don't have the key. Or you want to optionally fetch it, but only if it's unencrypted. Etc. Currently, there is no wallet utility that stores the AES key permanently. Therefore, I wrote it assuming the user might not have access to it.

src/crypto/signature.ts
RmbRT: Yeah I can do that later on. That was just the thing that took the least amount of effort and I was on a deadline. I didn't want to have to add more abstractions which would probably have broken things in other places. I didn't want to cause an even bigger refactoring.

src/crypto/wildcard/dh.ts
RmbRT: Yeah it's an artefact of cramming that into the signed message abstraction. It needs to be called that to fulfill the sigVerifier interface.

src/api/calls/getaccount.ts
DragonDev1906: (coming from Rust, where the data sent is usually a separate type)

So what do we do if we ever want to deserialize this type? It is not possible to get the DH secret key (as it's not in the serialized bytes) and you couldn't just set this to undefined, since re-serializing would then result in different bytes. And DHPair always has both sk and pk.

```ts
declare class DHKeyPair {
	privateKey: DHPrivKey;
	publicKey: DHPubKey;
}
```

So effectively the class/type definition states that you cannot expect the same bytes after deserializing and re-serializing this type if it `aes_secret !== undefined`.

That's why I'd argue that the in-memory and the to-be-serialized representation should be different types, unless we know for certain that we'll **never** need to deserialize this type (e.g. as part of a "give me the transaction with Hash X" request).

src/api/calls/getaccount.ts
DragonDev1906: Isn't that case already handled by providing `aes_secret`, which causes the enclave to return said AES key?

If the user has asked the enclave to encrypt its balance, why would he want to then ask it to give back the decrypted balance in plaintext? If that's the case you'd have to re-encrypt in the enclave and share the temporary key, so why not just use the AES key we've just returned. In that case the enclave always either returns encrypted (if an AES key is present and thus the operator doesn't know the balances) or in plaintext (if the operator already knows the balances).

At that point the only thing left is a boolean "do I need/want the balances" and the only place I can think of where that would be false is when fetching someone else's account for its address(es).

Or am I missing something?

src/crypto/signature.ts
DragonDev1906: For a future PR: https://github.com/perun-network/erdstall-ts-sdk/issues/163

src/crypto/wildcard/dh.ts
DragonDev1906: Then I'd suggest to add a comment to explain that, otherwise someone in the future will get quite confused why AES-GCM has something to do with signature verification.

src/api/calls/getaccount.ts
RmbRT: I want to keep it general for now. The exact design of the queries will probably change before mainnet anyway. And then we can see what is useful and what isn't. So for now, I wanted to allow querying different parts of the account separately, including sensitive information, whether that be plaintexts of private data, or the secret key for encryption via ECDH. And sometimes you may want to query some other person's account, and get info about it, such as the associated addresses & wildcard ID, but only query the balances if they are unencrypted. You can do all this with the current design.

src/api/responses/balanceproof.ts
RmbRT: because we already inherit from ErdstallObject.

src/api/calls/getaccount.ts
RmbRT: currently it just cannot deserialise this type. But since it's a query-type transaction anyway, I did not intend for a public receipt to be generated anyway. And direct receipts just contain the TX result & status code, not the TX itself.

package.json
RmbRT: Yeah I deleted 0.4.0 because I made a mistake during publishing.
I simply did not yet update the package version to 0.5.0 because I did not yet intend to publish the current state on npm until we have a working deployment.

src/api/calls/getaccount.ts
DragonDev1906: I still think it's overly general (or confusing):
User A has no business in seeing the balances of user B, even if the operator can see them and they're stored in plaintext, so when fetching someone else's account the operator should enforce that this is never sent (something this enum doesn't represent well.
And for your own account you either have the decryption key or you get it within the same response, which is what you want unless you tell the operator that you're not interested in your balance at all.

> allow querying different parts of the account separately

Still possible with a boolean that just states if you're interested in that data or not.

> but only query the balances if they are unencrypted

See my comment on why most users probably don't want that, though that check only works if the request is signed of course.

If we still want that, the current names are bad (sorry, but that's my opinion). I'm still not sure what each one actually means. Suggestion:
- true or "always": Always return the balances
- false or "never" or "null" or "undefined": Never return the balances
- "plaintext_only": Only return the balances if it isn't encrypted.

These clearly map to what the user wants, is descriptive and maps 1:1 to what the operator has, while the current names doen't really imply what the operator should do if it is encrypted and he gets "plain" [^1]: Should he return no balances, should he return null, should he ask the enclave for which a signature is required (causing unnecessary work since the operator already has the data encrypted for that user). If it is the last option that's extra implementation effort and resource consumption for the CoreTEE (at the very least in terms of bandwidth), something we should try to avoid.

Adding features in the future tends to be easier than removing them (which is always a breaking change).

[^1]: Unless I misunderstood what "plain" is supposed to mean.

src/api/calls/getaccount.ts
DragonDev1906: Isn't `SignedMessage` a newly added type?

src/api/responses/balanceproof.ts
DragonDev1906: Ah, I overlooked that `SignedMessage` doesn't inherit from `ErdstallObject`, that explains it of course.

src/api/calls/getaccount.ts
DragonDev1906: It's not just receipts, it could also be a subscription on all (plaintext) transactions that have been sent, if we want to add that at some point.

Are you sure we will never want to add code to save a transaction anywhere (e.g. browser local data) without needing a breaking change?

That we'll never want to add the corresponding de-serialize code without needing a breaking change (for example because someone wants to receive that type for some reason or store it somewhere)?

If that's the case we can leave it as is.

DragonDev1906: For now no change, may be done in the future.

src/api/calls/getaccount.ts
RmbRT: - [x] remove force_plaintext

RmbRT: Changed the parameter to:
 * true | "always": fetch regardless of encryption
 * false | undefined: do not fetch.
 * "only_if_plaintext": only fetch if not encrypted

src/crypto/wildcard/dh.ts
RmbRT: done

DragonDev1906 (approved):

